PR #1, opened 2015-01-12
JOHNZON-35

premsangeeth:
Mapper fails on short and byte array
Changed the use of Integer.class.cast to Number.class.cast


premsangeeth:
Merged

PR #7, opened 2017-05-07
Fix long number parse in collections, and unify non-integral parse

kumm:
On one code path non-integrals parsed as Double, on the other path
as BigDecimal.
Now it's always BigDecimal.

### DIFF
--- a/johnzon-jaxrs/src/main/java/org/apache/johnzon/jaxrs/ConfigurableJohnzonProvider.java
+++ b/johnzon-jaxrs/src/main/java/org/apache/johnzon/jaxrs/ConfigurableJohnzonProvider.java
@@ -224,4 +224,8 @@ public class ConfigurableJohnzonProvider<T> implements MessageBodyWriter<T>, Mes
             throw new IllegalArgumentException(e.getCause());
         }
     }
+
+    public void setUseBigDecimalForFloats(final boolean useBigDecimalForFloats) {
+        builder.setUseBigDecimalForFloats(useBigDecimalForFloats);
+    }
 }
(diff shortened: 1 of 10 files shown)


rmannibucau:
Hello @kumm , a few comments:

1. the map handling was intentional cause it doesn't target big numbers generally but only small ones (config). However we can enhance it a bit as you propose.
2. the convertTo new handling is however a regression in term of performance and behavior since double numbers become bigdecimal so you need to adapt. Don't forget to revert the deletion you did in the test, it is more important to support double than BigDecimal.

Overall I think the idea is good but ensure you use (in order) these types:

1. int
2. long
3. double
4. BigDecimal

cause it is the way it is casted in practise and rarely with a type check before (kind of implicit contract in term of usage I'd say).

Can you update this PR to support double please and add it back in the test?

kumm:
yes, I am going to update this.

kumm:
I'm not sure it's a good strategy to decide whether the JsonNumber should be a Double or BigDecimal.
It depends on use-case. 
Any feedback are welcome.

rmannibucau:
If you think about Map<String, ?> usage it is rarely for very big numbers, more for config and direct double casting is often done cause you know what kind of number you get. If you have another proposal I'm happy to discuss it (I'm not 100% happy too with this stratégy).

kumm:
What about a MapperConfig setting?
Something like USE_BIG_DECIMAL_FOR_FLOATS here: https://github.com/FasterXML/jackson-databind/wiki/Deserialization-Features

rmannibucau:
sounds good to me

kumm:
should I implement it as a new PR (new story), on top of this, or just update this one?

rmannibucau:
there is fine, guess you will just replace some existing conditions by the new flag

kumm:
missed to comment after update, but it's updated

rmannibucau:
Looks almost good for me, only missing part if the wiring of that config in the exposed object like *Provider (JAXRS providers), can you add it (just a setter returning void generally). Except that it is a great patch!

kumm:
Thanks, done.
Just a question: setSerializeValueFilter intentionally left out from WildcardConfigurableJohnzonProvider?

rmannibucau:
Dont think so, pby a miss

rmannibucau:
applied, thanks a lot @kumm

kumm:
It was a pleasure.
Closing it.

PR #8, opened 2017-06-24
JOHNZON-126: Rename JohnsonJsonb to JohnzonJsonb

jwcarman:


jeanouii:
It was applied and fixed. Closing

PR #10, opened 2017-06-24
JOHNZON-127: Make CDI Dependencies Optional in OSGi

jwcarman:


jeanouii:
Also applied already. Closing

PR #13, opened 2017-10-14
Add support for converting java.time.Instant to JSON string and vice versa

bcluap:
Includes new converter with test cases.

Will parse JSON strings into an Instant so long as they are in a ISO_8601 date+time compliant format
e.g.:
2007-12-03T14:15:30.00Z
2007-12-03T17:15:30+03:00
2007-12-03T08:15:30-06:00
2007-12-03T08:15:30.000-06:00

Instant parsed to String will be in ISO_8601 format and have the local system timezone in the representation:
e.g. Say the system timezone is +02:00 and the Instant is 2007-12-03T14:15:30.00Z then the resulting String would be 2007-12-03T16:15:30+02:00

### DIFF
--- a/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MapperBuilder.java
+++ b/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MapperBuilder.java
@@ -56,6 +56,7 @@ import java.math.BigInteger;
 import java.net.URI;
 import java.net.URL;
 import java.nio.charset.Charset;
+import java.time.Instant;
 import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


rmannibucau:
Hi @bcluap ,

what about just importing jsonb converters instead of creating another "branch" and 2 parallel code bases to maintain?

-> https://github.com/apache/johnzon/blob/master/johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JohnzonBuilder.java#L439

bcluap:
Will create new pull request incorporating jsonb builders

PR #5, opened 2017-03-13
JOHNZON-108 Add dynamic ServiceLoader support for OSGi environments

stefanseifert:

PR #15, opened 2017-10-23
Rework of JSON-B Type Converters to use Mapper Converters and add Java8 Date/Time Converters to Mapper Project

bcluap:
Includes new converters with test cases.

Will parse JSON strings into an Instant/Date so long as they are in a format that can be heuristically determined.
e.g. these are all includes in test cases:
2007-12-03T11:15:00-03:00
2007-12-03T14:15:00Z
2007-12-03T17:15:00+0300
2007-12-03T17:15:00.0+03:00
2007-12-03T17:15:00.0+0300
2007-12-03T17:15:00.00+03:00
2007-12-03T11:15:00.00-03:00
2007-12-03T17:15:00.00+0300
2007-12-03T17:15:00.000+03:00
2007-12-03T17:15:00.000+0300
2007-12-03T14:15:00.000+0000
2007-12-03T14:15:00.000Z
2007-12-03T17:15+03:00
2007-12-03T17:15+0300
2007-12-03T14:15:00
2007-12-03T14:15Z
2007-12-03T14:15
20071203141500Z
200712031415Z
20071203111500-0300
20071203171500+0300
20071203111500-03:00
20071203171500+03:00
200712031115-0300
200712031115-03:00
200712031415
20071203141500

Instant parsed to String will be in ISO_8601 format with UTC (Z) timezone.
Date parsed to String will be in format yyyyMMddHHmmssZ with UTC (Z) timezone so as to be backwards compatible.

### DIFF
--- a/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/converter/ZonedDateTimeConverter.java
+++ b/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/converter/ZonedDateTimeConverter.java
@@ -25,8 +25,7 @@ import static org.apache.johnzon.mapper.converter.Java8Converter.ZONE_ID_UTC;
 
 public class ZonedDateTimeConverter extends Java8Converter<ZonedDateTime> {
 
-    
-    private static final DateTimeFormatter FORMATTER_OUT = DateTimeFormatter.ISO_OFFSET_DATE_TIME.withZone(ZONE_ID_UTC);
+    private static final DateTimeFormatter FORMATTER_OUT = DateTimeFormatter.ISO_OFFSET_DATE_TIME;
     // 2007-12-03T17:15:00+03:00
     // 2007-12-03T17:15:00.0+03:00
     // 2007-12-03T17:15:00.00+03:00
@@ -76,11 +75,13 @@ public class ZonedDateTimeConverter extends Java8Converter<ZonedDateTime> {
 
     // 200712031115-03:00
     private static final DateTimeFormatter FORMATTER_IN_PATTERN_14 = DateTimeFormatter.ofPattern("yyyyMMddHHmmXXX");
-    
+
     // 2007-12-03T14:15:00
     private static final DateTimeFormatter FORMATTER_IN_PATTERN_15 = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss").withZone(ZONE_ID_UTC);
-    
-    
+
+    //2007-12-03T14:15:00.0Z
+    private static final DateTimeFormatter FORMATTER_IN_PATTERN_16 = DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss.SX");
+
     /**
      * Returns the Instant in format yyyy-MM-dd'T'HH:mm:ssZ
      * @param zdt
@@ -88,7 +89,7 @@ public class ZonedDateTimeConverter extends Java8Converter<ZonedDateTime> {
      */
     @Override
     public String toString(final ZonedDateTime zdt) {
-        return zdt.toString();
+        return FORMATTER_OUT.format(zdt);
     }
 
     /**
@@ -129,8 +130,12 @@ public class ZonedDateTimeConverter extends Java8Converter<ZonedDateTime> {
                         // 2007-12-03T17:15+0300
                         return ZonedDateTime.from(FORMATTER_IN_PATTERN_12.parse(text));
                     case 22:
-                        // 2007-12-03T17:15+03:00
-                        return ZonedDateTime.from(FORMATTER_IN_PATTERN_2.parse(text));
+                        // 2007-12-03T17:15+03:00 or 2018-02-09T10:32:01.1Z
+                        if (text.endsWith("Z")) {
+                            return ZonedDateTime.from(FORMATTER_IN_PATTERN_16.parse(text));
+                        } else {
+                            return ZonedDateTime.from(FORMATTER_IN_PATTERN_2.parse(text));
+                        }
                     case 24:
                         if (text.endsWith("Z")) {
                             // 2007-12-03T14:15:00.000Z
@@ -199,4 +204,3 @@ public class ZonedDateTimeConverter extends Java8Converter<ZonedDateTime> {
         }
     }
 }
-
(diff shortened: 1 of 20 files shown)


bcluap:
Any chance of this being included. Been using the patch for a while and it works perfectly

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/converter/ZonedDateTimeConverter.java
```diff
@@ -0,0 +1,202 @@
...
+ * under the License.
+ */
+package org.apache.johnzon.mapper.converter;
+
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
+import static org.apache.johnzon.mapper.converter.Java8Converter.ZONE_ID_UTC;
+
+public class ZonedDateTimeConverter extends Java8Converter<ZonedDateTime> {
```
derekm: @bcluap -- check this out: https://github.com/derekm/jersey/blob/fix-httpdateformat/core-common/src/main/java/org/glassfish/jersey/message/internal/HttpDateFormat.java#L62

https://github.com/derekm/jersey/blob/fix-httpdateformat/core-common/src/test/java/org/glassfish/jersey/message/internal/HttpDateFormatTest.java

I came up with the algorithm in readDate() and the tests, if you want to use them!

bcluap:
@derekm 

Thanks for your input. I was wondering about an approach of looping through different formats looking for the best match but my concern was the performance implications, I did a quick test now over 50M iterations of parsing an ISO 8601 date string (the most common format) with the 2 different implementations and the approach of testing each format for the best match is 10X slower. I have no doubt that my implementation is a bit hacky with all the string analysis (not proud of the code :-) ) but I coded it for raw speed due to the frequency of it being used.

derekm:
Yeah, I see... Your algorithm takes a sort of branch-and-bound strategy by examining the input string for characteristics unique to a group of formats. I could probably add a similar pre-processing step that turn date patterns on & off. Thanks for the input & benchmark!

derekm:
@bcluap -- Your previous performance tests demonstrate the fact that my naive approach tests on average 10 patterns before attempting any ISO 8601-related patterns.

As your latest commit demonstrates, supporting extensions or additions to the date patterns is an important feature of a utility library such as this.

What if the patterns could be defined with a set of predicates that test for whether the pattern should be included in parsing attempts?

If each pattern could provide its own set of pre-processing predicates, could the "Is 'T' the eleventh character?" test become shared somehow so it is only executed once for all patterns with that predicate? ... Even if it couldn't initially, might repeated execution of predicates be a worthwhile trade-off for extensibility vs. minimizing number of operations?

bcluap:
Hi Derek,

Yea I get you - some sort char streaming where a parser works through the
string one char at a time passing each to a set of patterns that are
"listeners"  and opt-out as they see that their predicates are not met.
Then by the end of the string the patterns remaining are run till one
replies without an error.

I guess its kind of like my logic but in a more modular fashion where each
pattern decides its criteria as opposed to a "big brother" logic tree that
has knowledge of all the patterns.

I'll see what I can come up with. Pretty busy at the moment getting a new
company off the ground but will find some time. What are the timelines
looking like for getting a new 7.0.5 TomEE build out with the Johnzon
changes - I'm running my own Johnzon fork right now and its getting messy
:-)

Cheers

Paul

derekm:
Paul @bcluap -

I'm not affiliated with TomEE. I accident'd across this PR because I broke your TomEE server's Swagger 2.0 / OpenAPI 3.0 integration here: https://github.com/swagger-api/swagger-core/pull/2627

I took an interest in this PR because I've jumped down this rabbit hole with respect to Jersey's content type negotiation & date parsing routines since 2014, and my work there is yet to be accepted (in part due to my own availability in fulfilling commitments following code reviews & obvious changes) ... and now Jersey is being transferred to the Eclipse Foundation, pausing development further as I'm finally getting back into it!

If we could combine our approaches into a library that both Johnzon & Jersey could use, that might be an amazing feat of engineering! (If the pre-processor predicates that test positive select more than one pattern, you'll want my selection-of-successful-parsings algorithm from my Jersey code! And maybe some of the patterns and tests, too.)

Re: maintaining private forks, I feel your pain! I'm maintaining a private SNAPSHOT build of Jersey with improvements to its proxy client (an early predecessor to MicroProfile Rest Client 1.0 that actually had little to no influence on the MP Rest Client spec [MP Rest Client authors didn't know about `jersey-proxy-client` until late in spec writing]).

I have been working with @johnament on @hammock-project lately. I know John is a Johnzon (fittingly), but I don't know how close to TomEE contacts he is. Perhaps I'll help by trawling twitter for a TomEE/Johnzon authority...

Cheers,

Derek

derekm:
PS: Here is my previous work in Jersey on this topic:

https://github.com/jersey/jersey/issues/2814

https://github.com/jersey/jersey/pull/94

struberg:
Hi folks!

Thanks a lot for your feedback! 
github pull requests are - for a good reason - not the communication of choice at ASF projects.
I personally have seen it way too many conflicts with ambugious ticket numbers, e.g. #5 which exists in dozen downstream repos with different content. And there are dozen more issues with the simple github ticket system for projects at ASF scale.

Of course we love all the discussions and I'd like to pull them over to our official issue tracker 
and mailing lists. http://johnzon.apache.org/mail-lists.html
https://issues.apache.org/jira/browse/JOHNZON

I'll gradually work through the open PRs and pull them over. 
Happy to see you over there!

txs and LieGrue,
strub

struberg:
I came back to close up all the PRs and there is good news: support for time types is already available in Johnzon because it's required by the JSON-B spec even.
So I think we can consider this PR closed - txs!

bcluap:
I am correct in saying that TomEE 7.0.5 and hence Johnzon 1.0.1 still does not support Java 8 times?

Johnzon serialised a ZonedDateTime into something like:

{"dateTime":{"date":{"dayOfWeek":"SUNDAY","month":"APRIL","year":2018,"dayOfMonth":15,"dayOfYear":105,"era":"CE","monthValue":4,"chronology":{"calendarType":"iso8601","id":"ISO"},"leapYear":false,"day":15},"dayOfWeek":"SUNDAY","hour":13,"month":"APRIL","dayOfMonth":15,"dayOfYear":105,"year":2018,"monthValue":4,"nano":0,"time":{"hour":13,"nano":0,"minute":32,"second":13},"minute":32,"second":13},"offset":{"totalSeconds":0,"rules":{"standardOffsets":[{"totalSeconds":0,"rules":{"standardOffsets":[{"totalSeconds":0,"rules":{"standardOffsets":....goes on for pages

struberg:
hi @bcluap yes Johnzon-1.0.1 does not support it, but if you upgrade to TomEE-8 (JavaEE 8 level) then it will work out of the box.

PR #18, opened 2018-03-27
JsonbConfig injectable by JAX-RS Provider 

amoscatelli:


rmannibucau:
Hi @amoscatelli,

1. can you fix the formatting diff please to ensure the patch is "clean"?
2. can you just hardcode a type in https://github.com/apache/johnzon/pull/18/files#diff-173d2c457f1e7c92c8500cb2d20edf0eR166 since the type is not repsected for the resolution (the provider will handle all json types but the init will only provide a single one, Object.class can be enough when you do getContext). Alternative is to not cache (delegate) the jsonb instance but it would slow down the runtime too much to be a valid default IMHO.
3. I really the like the test but is it possible to not add the provider by default to ensure other tests don't use that mecanism (to cover the else codepath)?

Otherwise it looks like a very good - and relevant - addition

Romain

amoscatelli:
OK for point 1 and point 3 : done !

I do not agree on point 2 :
Adherence to the specification should be more important. getContext expects the type/class and, to increase the reusability of this class, I think we should keep on passing the actual type/class. We don't know what kind of logic could be used by others or in future to return a Jsonb from a Provider depending on the passed type. So I would not hardcode it for the sake of flexibility. 

I'll try to rewrite the code to make both of us happy.
Give me a couple of minutes.

rmannibucau:
@amoscatelli: agree we should respect the spec but the spec doesn't define what the type is used for. Concretely it is used to get an instance of the "context" relevant for the type. In your case whatever type you have you will use the same instance. An alternative is the following one:

1. get rid of the atomicref to use a Supplier<Jsonb>
2. at first request check if the contexprovider gives an instance, if not -> set a "constant" supplier ( () -> instance ), if so just use the context provider instance ( () -> providers.get(....).getContext(type) ) without any cache

this keeps the same default and aligns on the context provider API

amoscatelli:
I agree now, I was already doing something similar. Commit incoming.

amoscatelli:
@rmannibucau here we are !

amoscatelli:
I made one final fix to keep all of the old methods inside the test class

rmannibucau:
Applied - think just before the small tweaks

Created https://issues.apache.org/jira/browse/JOHNZON-163 and updated the jsonb instance handling to ensure we don't keep recreating it which is a perf killer.

amoscatelli:
@rmannibucau  Should I close the pull request ? What's the procedure ?

rmannibucau:
@amoscatelli yes please (I dont have the perms to do so ATM)

amoscatelli:
@rmannibucau in a couple of hours I shall be making a new pull request with other set of changes. Thank you for your support.

PR #19, opened 2018-03-30
complete support for DateFormat

amoscatelli:


amoscatelli:
I noticed that DateFormat specified in JsonbConfig were not used to serialize and deserialize properly.
I also made a proper teste specified for the Date types.

PR #20, opened 2018-04-03
deserialization of enum from deserializationContext (normally used inside jsonbDeserializers)

amoscatelli:
Deserialization of enum from deserializationContext (normally used inside jsonbDeserializers).
This pull also contains a partial refactor for Mappings to manage TypeVariable and ParameterizedType.
I also fixed milli, micro, nano seconds formatting for date/temporal accessor types being truncated.

### DIFF
--- a/johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
+++ b/johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
@@ -585,6 +585,8 @@ public class JsonbAccessMode implements AccessMode, Closeable {
         }
         final Type[] genericInterfaces = aClass.getGenericInterfaces();
         return Stream.of(genericInterfaces).filter(ParameterizedType.class::isInstance)
+                .filter(i -> ParameterizedType.class.cast(i).getRawType() instanceof Class)
+                .filter(i -> ParameterizedType.class.cast(i).getActualTypeArguments()[0] instanceof Class)
                 .filter(i -> Adapter.class.isAssignableFrom(Class.class.cast(ParameterizedType.class.cast(i).getRawType())))
                 .findFirst()
                 .map(pt -> payloadType.isAssignableFrom(Class.class.cast(ParameterizedType.class.cast(pt).getActualTypeArguments()[0])))
--- a/johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
+++ b/johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
@@ -96,6 +96,7 @@ import java.util.function.Supplier;
 import java.util.stream.Stream;
 
 import static java.util.Arrays.asList;
+import java.util.Objects;
 import static java.util.Optional.ofNullable;
 import static org.apache.johnzon.mapper.reflection.Converters.matches;
 
@@ -586,6 +587,7 @@ public class JsonbAccessMode implements AccessMode, Closeable {
         final Type[] genericInterfaces = aClass.getGenericInterfaces();
         return Stream.of(genericInterfaces).filter(ParameterizedType.class::isInstance)
                 .filter(i -> ParameterizedType.class.cast(i).getRawType() instanceof Class)
+                .filter(i -> Objects.nonNull(ParameterizedType.class.cast(i).getActualTypeArguments()) && ParameterizedType.class.cast(i).getActualTypeArguments().length>0)
                 .filter(i -> ParameterizedType.class.cast(i).getActualTypeArguments()[0] instanceof Class)
                 .filter(i -> Adapter.class.isAssignableFrom(Class.class.cast(ParameterizedType.class.cast(i).getRawType())))
                 .findFirst()
--- a/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mapper.java
+++ b/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mapper.java
@@ -140,7 +140,7 @@ public class Mapper implements Closeable {
     private boolean isDeduplicateObjects(Class<?> rootType) {
         Boolean dedup = config.isDeduplicateObjects();
         if (dedup == null) {
-            Mappings.TypeMapping classMapping = mappings.findOrCreateTypeMapping(rootType);
+            Mappings.TypeMapping classMapping = mappings.findOrCreateTypeMapping(rootType, mappings.getApplyObjectConverter(rootType));
             if (classMapping != null && MappingType.CLASSMAPPING.equals(classMapping.getType())) {
                 dedup = classMapping.as(ClassMapping.class).isDeduplicateObjects();
             }
--- a/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mapper.java
+++ b/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mapper.java
@@ -44,6 +44,8 @@ import java.util.ArrayList;
 import java.util.Collection;
 
 import static java.util.Arrays.asList;
+import org.apache.johnzon.mapper.Mappings.ClassMapping;
+import org.apache.johnzon.mapper.Mappings.MappingType;
 import static org.apache.johnzon.mapper.internal.Streams.noClose;
 
 public class Mapper implements Closeable {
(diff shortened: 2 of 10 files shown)


johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
```diff
@@ -585,6 +585,8 @@ private boolean isReversedAdapter(final Class<?> payloadType, final Class<?> aCl
         }
         final Type[] genericInterfaces = aClass.getGenericInterfaces();
         return Stream.of(genericInterfaces).filter(ParameterizedType.class::isInstance)
+                .filter(i -> ParameterizedType.class.cast(i).getRawType() instanceof Class)
+                .filter(i -> ParameterizedType.class.cast(i).getActualTypeArguments()[0] instanceof Class)
```
rmannibucau: maybe check the length too in this lambda, can fail otherwise ;)

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mapper.java
```diff
@@ -138,9 +140,9 @@ public void writeObject(final Object object, final Writer stream) {
     private boolean isDeduplicateObjects(Class<?> rootType) {
         Boolean dedup = config.isDeduplicateObjects();
         if (dedup == null) {
-            Mappings.ClassMapping classMapping = mappings.findOrCreateClassMapping(rootType);
-            if (classMapping != null) {
-                dedup = classMapping.isDeduplicateObjects();
+            Mappings.TypeMapping classMapping = mappings.findOrCreateTypeMapping(rootType);
+            if (classMapping != null && MappingType.CLASSMAPPING.equals(classMapping.getType())) {
```
rmannibucau: just a style thing but maybe add in TypeMapping a as(Class<T>) to be able to do

    typeMapping.as(ClassMapping.class)

it avoids to put cast everywhere and makes the pattern if (type == xxx) tm.as(X.class) more natural IMHO

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mappings.java
```diff
@@ -98,8 +119,47 @@ public Boolean isDeduplicateObjects() {
...
         }
 
+        @Override
+        public MappingType getType() {
+            return MappingType.CLASSMAPPING;
+        }
+
     }
 
+    public static class ParameterizedTypeMapping extends TypeMapping {
```
rmannibucau: this class should be able to provides its parameters as TypeMapping to not require reflection at runtime no?

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mappings.java
```diff
@@ -98,8 +119,47 @@ public Boolean isDeduplicateObjects() {
...
+        }
+
+        @Override
+        public MappingType getType() {
+            return MappingType.PARAMETERIZEDTYPEMAPPING;
+        }
+
+    }
+
+    public static class TypeVariableMapping extends TypeMapping {
```
rmannibucau: same than for parameterized type, we should cache bounds probably as TypeMapping (ClassMapping in the best case), no?

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
amoscatelli: done

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mapper.java
amoscatelli: done

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mappings.java
amoscatelli: what ? Isn't this what I am actually doing ?

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mappings.java
amoscatelli: again, isn't this what I am already doing ?
In Mappings.class I add both ParameterizedTypeMapping or TypeVariableMapping to the cache, with classes.putIfAbsent

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mappings.java
rmannibucau: Not really cause it means the instances of done at https://github.com/apache/johnzon/pull/20/files#diff-7f654235e6be2d22d4eb2df20cd40b19R220 could be replaced by a findTypeMapping directly (then a switch on the type and the new TypeMapping hierarchy should allow to align all the implementations with our old CollectionMapping (https://github.com/amoscatelli/johnzon/blob/bfbed85a045bcb53c071e77d01a1b52ca29d6c70/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java#L187) instead of creating a hierarchy we actually don't use.

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mapper.java
```diff
@@ -138,10 +142,15 @@ public void writeObject(final Object object, final Writer stream) {
     private boolean isDeduplicateObjects(Class<?> rootType) {
         Boolean dedup = config.isDeduplicateObjects();
         if (dedup == null) {
-            Mappings.ClassMapping classMapping = mappings.findOrCreateClassMapping(rootType);
-            if (classMapping != null) {
-                dedup = classMapping.isDeduplicateObjects();
+            if (!deduplicateMap.containsKey(rootType)){
+                JohnzonDeduplicateObjects jdo = rootType.getAnnotation(JohnzonDeduplicateObjects.class);
```
rmannibucau: this should probably move to ClassMapping no?

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mappings.java
```diff
@@ -332,39 +329,50 @@ public static boolean isPrimitive(final Type type) {
...
+        TypeMapping classMapping = classes.get(clazz);
         if (classMapping == null) {
-            if (!Class.class.isInstance(clazz) || Map.class.isAssignableFrom(Class.class.cast(clazz))) {
-                return null;
+            if (Class.class.isInstance(clazz)) {
+                if (!Map.class.isAssignableFrom(Class.class.cast(clazz))){
+                    classMapping = createTypeMapping(Class.class.cast(clazz), applyObjectConverter);
+                }
+            } else if (ParameterizedType.class.isInstance(clazz)){
+                classMapping = findOrCreateTypeMapping(ParameterizedType.class.cast(clazz).getRawType(), applyObjectConverter);
```
rmannibucau: I'd add the Stream.of(pt.getActualParameters()).map(it -> findOrCreateTypeMapping(it).collect(toList()) in the model + probably the CollectionMapping meta here to "normalize" the whole model and make it consistent

side note: take care for types which look like Foo<? extends Foo> and avoid loops (some tests to add pby)

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mappings.java
```diff
@@ -332,39 +329,50 @@ public static boolean isPrimitive(final Type type) {
...
+            } else if (TypeVariable.class.isInstance(clazz)){
+                TypeVariable cast = TypeVariable.class.cast(clazz);
+                if (Objects.nonNull(cast.getAnnotatedBounds()) && cast.getAnnotatedBounds().length > 0){
+                    classMapping = findOrCreateTypeMapping(TypeVariable.class.cast(clazz).getAnnotatedBounds()[0].getType(), applyObjectConverter);
```
rmannibucau: probably store it in the TypeVariableMapping too?

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mapper.java
amoscatelli: Why?

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mappings.java
amoscatelli: I removed multiple type of mapping after What we said. Do we need them?

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mappings.java
amoscatelli: What model would you add the list to? Do I have to restore ParemeterizedTypeMapping?

PR #27, opened 2018-09-28
JOHNZON-189 Add support for JAX RS Whiteboard

rotty3000:


johnzon-jsonb/src/main/java/org/apache/johnzon/jaxrs/jsonb/jaxrs/JsonbJaxrsProvider.java
```diff
@@ -54,9 +54,10 @@
 @Provider
 @Produces(MediaType.APPLICATION_JSON)
 @Consumes(MediaType.APPLICATION_JSON)
-@Priority(value = 4900)
+@Priority(value = JsonbJaxrsProvider.PRIORITY)
 public class JsonbJaxrsProvider<T> implements MessageBodyWriter<T>, MessageBodyReader<T>, AutoCloseable {
 
+    public static final int PRIORITY = 4900;
```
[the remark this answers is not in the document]
rotty3000: I wasn't even sure if this was really that interesting.. I just didn't want to ignore the behaviour. I can just remove it outright and assign the service it's own service ranking. And sure, perhaps the osgi side can read the annotation. It's just not likely that important to have exactly the same value.

johnzon-jsonb/src/main/java/org/apache/johnzon/jaxrs/jsonb/jaxrs/osgi/Activator.java
```diff
@@ -0,0 +1,150 @@
...
+import org.osgi.framework.BundleActivator;
+import org.osgi.framework.BundleContext;
+import org.osgi.framework.Constants;
+import org.osgi.framework.PrototypeServiceFactory;
+import org.osgi.framework.ServiceRegistration;
+import org.osgi.service.jaxrs.whiteboard.JaxrsWhiteboardConstants;
+import org.osgi.util.converter.Converter;
+import org.osgi.util.converter.Converters;
+
+public class Activator implements BundleActivator {
```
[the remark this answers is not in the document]
rotty3000: > should we create an osgi module or create 2 deliveries (with an osgi classifier): one with these classes and one without?

if that's what you want, we can do that. I didn't see the need.

> Scanners can fail on these classes, not sure we want to bundle them like that directly.

How so? Perhaps I've missed something.

> It looks also risky to have OSGi as dependencies and it can be users in other modules unexpectedly no?

All the osgi dependencies are `provided`. Everything is internal to the bundle. Not sure how others would be affected by this. Only when run in OSGi does this become relevant.

johnzon-jsonb/src/main/java/org/apache/johnzon/jaxrs/jsonb/jaxrs/JsonbJaxrsProvider.java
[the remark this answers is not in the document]
rotty3000: You were right that it was easier through reflection. I pushed an update.

rotty3000:
some updates to isolate the OSGi stuff a little better
- make sure the package is not exported
- make as many types package private as possible

rotty3000:
@rmannibucau so does that mean you want to remove all the OSGi details from the original artifacts?

rmannibucau:
The point is to not take any risk and break some env - log4j broke some just adding a module-info.java :s. If not too costly Id be for it - probably using reflection to activate all we can with a single osgi module. If too much code let's go with a classifier trick short term.

rotty3000:
> The point is to not take any risk and break some env

of course.

> If not too costly Id be for it - probably using reflection to activate all we can with a single osgi module.

Why would we need reflection?

rotty3000:
I guess the question is do you want to make an OSGi module that encompasses all the parts together in one jar (which would mean that maybe you could remove the osgi metadata from the original jars, but that would break all users who already use them)?

OR, a slim osgi jar that is installed along _with_ the others and simply boots the parts missing for deeper OSGi integration.

rotty3000:
the one I could live with the most is a bootstrap module per feature. a.k.a. proliferation of modules.

on the other hand maven classifiers are just really painful :( they live in the realm of "mutliple artifacts per project" which is a huge maven anti-pattern and Sonatype even recommends against it.

rotty3000:
One last comment is that I'm trying to make Apache Johnzon the poster child for a regular java library which also has awesome, modern OSGi integration. This set of changes is only one step away from that. The upcoming bnd version (and maven-bundle-plugin) will help complete that.

rmannibucau:
Do you mean next bundle plugin does not need that code at all? Can it generate one jar per module with osgi classifier with the required part per module + its feature.xml? Would be the best of all no?

rotty3000:
> Do you mean next bundle plugin does not need that code at all?

Oh no, that code is needed. But next bnd has new bundle annotations which permit eliminating a lot of typing metadata in the pom file, helping to calculate it during build time.

> Can it generate one jar per module with osgi classifier with the required part per module + its feature.xml? Would be the best of all no?

bnd is not about creating more artifacts per project. I also would not use classifiers. I would simply create additional projects, something like `johnzon-jaxrs-osgi` which imports packages and classes exported from `johnzon-jaxrs`. Likewise with `johnzon-jsonb` (i.e. `johnzon-jsonb-osgi`). At least this way the original jars stay like they were, but the OSGi resolver can figure out the relationship and assemble a reliable runtime.

rmannibucau:
Hmm, here are my current thoughts:

1. most of the OSGi classes are generable - so we can write a mojo or temporarly a .groovy script
2. thanks to 1 i wouldnt create a new module but just a new artifact with a osgi classifier which would create johnzon-jsonb-xxxx-osgi which is smoother as convention since automatic and therefore predictable (vs distinct module which is not)

If it can wait 2 weeks i can try to PoC 1. AFAIK it just requires some constant definition and everything else can be scanned, no?

rotty3000:
I don't see how the code can be generated. There's specific details specific to each case. I really don't like the idea of creating classifier artifacts. It's going to produce complications as it always does. If you're going to go to all that trouble, why not just create two new projects & artifacts completely?

rotty3000:
Hey guys, here's a new implementation of this, no new classes, all annotations based!

rotty3000:
This was tested with https://github.com/apache/aries-cdi

rmannibucau:
Hi @rotty3000, is cdi optional in osgi with the requirecdiextender? It should stay optional since everything is usable without cdi. Also wonder if all the provider must be discovered, mapper and jsonb modules are deployed together and users tend to use configurable flavor quite often so should they immediately use ranking and priority?

Otherwise +1 to move to annotations.
Open point: dedicated module or not? Think it can make sense to be able to use ConfigAdmin to switch to configurable providers but want to get your opinion on that.

rotty3000:
@rmannibucau yeah, I forgot to make the CDI requirements optional, Doing that!

rotty3000:
As for the configuration, we'll need much more code for that :(

My thought process on this was: If you would need much finer configuration, unless Johnzon project itself introduce something like DS, CDI or low level OSGi code, then the developer just do this like they would in other code as now.

However, for the common cases where all you need is what stock `org.apache.johnzon.jaxrs.JsrProvider` + `org.apache.johnzon.jaxrs.JohnzonProvider` provide then you can use johnzon-* as OSGi bundle libraries.

WDYT?

rmannibucau:
@rotty3000 i guess these days - ie since JSON-B is there - we mainly expect JsonbJaxrsProvider [1] to be the one used and no more JohnzonProvider (and JsrProvider is embedded in JsonbJaxrsProvider), means we have the choice between these options:

1. do an actual OSGi bundle/new module supporting optionally ConfigAdmin
2. only osgi-fy johnzon-jaxrs
3. only osgi-fy johnzon-jsonb
4. osgi-fy all of them

While SE and EE code has no impact at all - manifest changes are ignored by EE, no new transitive dependency, no new META-INF/services/xxx registration etc... - then I'm fine with any first step but as an user I'm expecting 1, in particular to configure ignorable types, be able to enable pretty mode etc...

All that to say I'm quite mixed since I see values in both 1 and you already done work and that while you can harness the EE case - guess meecrowave is a trivial test you can run - I'm happy to get any first step forward.

[1] https://github.com/apache/johnzon/blob/master/johnzon-jsonb/src/main/java/org/apache/johnzon/jaxrs/jsonb/jaxrs/JsonbJaxrsProvider.java#L68

rotty3000:
@rmannibucau what do you think of this now? This works really well. I would prefer to keep it modularized.

rotty3000:
I even added the new Aries CDI feature for the implicit extension so that when the Jaxrs part is used in CDI it would automatically get applied.

johnzon-osgi/src/main/java/org/apache/johnzon/osgi/Activator.java
```diff
@@ -0,0 +1,162 @@
...
+            putIfAbsent(
+                JaxrsWhiteboardConstants.JAX_RS_MEDIA_TYPE,
+                config.osgi_jaxrs_media_type());
+            putIfAbsent(
+                Constants.SERVICE_RANKING,
+                JsonbJaxrsProvider.class.getAnnotation(Priority.class).value());
+        }};
+    }
+
+    private class JsonbJaxrsProviderFactory implements PrototypeServiceFactory<JsonbJaxrsProvider<?>> {
```
rmannibucau: Should be static?

johnzon-osgi/src/main/java/org/apache/johnzon/osgi/Activator.java
```diff
@@ -0,0 +1,162 @@
...
+        }
+
+        @Override
+        public void ungetService(
+            Bundle bundle, ServiceRegistration<JsonbJaxrsProvider<?>> registration, JsonbJaxrsProvider<?> service) {
+        }
+
+    }
+
+    private class ExtendedJsonbJaxrsProvider extends JsonbJaxrsProvider<Object> {
```
rmannibucau: Static?

johnzon-osgi/src/main/java/org/apache/johnzon/osgi/Config.java
```diff
@@ -0,0 +1,43 @@
...
+
+import org.osgi.service.metatype.annotations.AttributeDefinition;
+import org.osgi.service.metatype.annotations.ObjectClassDefinition;
+
+@ObjectClassDefinition(
+    description = "Configuration for Johnzon JSONB JAX-RS Whiteboard Message Body Reader & Writer",
+    name = "Johnzon JAX-RS JSONB",
+    pid = Config.CONFIG_PID
+)
+@interface Config {
```
rmannibucau: public? (to avoid unsafe requirement and some surprises in some env, not strictly required but saner IMHO)

Also do we want all the provider config there?

johnzon-osgi/pom.xml
```diff
@@ -0,0 +1,131 @@
...
+    <version>1.2.3-SNAPSHOT</version>
+  </parent>
+  <modelVersion>4.0.0</modelVersion>
+
+  <artifactId>johnzon-osgi</artifactId>
+  <name>Johnzon :: Support for OSGI Jaxrs Whiteboard</name>
+  <packaging>bundle</packaging>
+
+  <dependencies>
+    <dependency>
```
rmannibucau: Do we need that? I d like to limit deps as much as possible and needs look simple enough here to not need it, wdyt?

johnzon-osgi/src/main/java/org/apache/johnzon/osgi/cdi/RegisterCdiExtension.java
```diff
@@ -0,0 +1,74 @@
...
+        } catch (ClassNotFoundException cfne) {
+            return false;
+        }
+    }
+
+    @SuppressWarnings("serial")
+    private static Map<String, Object> getCdiExtensionProperties() {
+        return new HashMap<String, Object>() {{
+            put(CDIConstants.CDI_EXTENSION_PROPERTY, "JavaJSONB");
+            put("aries.cdi.extension.mode", "implicit");
```
rmannibucau: :heart_eyes:

johnzon-osgi/src/main/java/org/apache/johnzon/osgi/Config.java
rotty3000: I added more of the config options. Some you'd have to define an encoding for the `Map<String,String>` input since you can't do that on an annotation.

johnzon-osgi/pom.xml
rotty3000: This is way more complex than I want to write. This logic is dynamic in that it will update as configurations change, create multiple instances if factory configurations are used... I don't want to write that.

rotty3000:
@rmannibucau where is the little doc you speak of?

rmannibucau:
@rotty3000 here https://github.com/apache/johnzon/tree/master/src/site/markdown

rotty3000:
some docs added.

rmannibucau:
Thanks @rotty3000 , validated it does not impact ee case and merged. A potential enhancement can be to add an IT to avoid us to break it unintentionally but this is already an awesome feature in, congrats!

PR #29, opened 2018-10-05
JOHNZON-192 Switch johnzon-maven-plugin class generation from manual source writer to a more robust solution

simonetripodi:


jeanouii:
Left a comment on JIRA. How should we move with this?

PR #38, opened 2019-03-31
JOHNZON-205: add support for arrays in toStructure

elexx:
fixes [JOHNZON-205](https://issues.apache.org/jira/browse/JOHNZON-205)

### DIFF
--- a/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mapper.java
+++ b/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mapper.java
@@ -39,7 +39,7 @@ import java.nio.charset.Charset;
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.List;
+
 import javax.json.JsonArray;
 import javax.json.JsonBuilderFactory;
 import javax.json.JsonObject;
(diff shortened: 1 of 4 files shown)


johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mapper.java
```diff
@@ -146,12 +146,31 @@ public JsonValue toStructure(final Object object) {
...
+            writeObject(objectToList(object), objectGenerator, null,
+                    isDeduplicateObjects(object.getClass()) ? new JsonPointerTracker(null, "/") : null);
+        } else {
+            writeObject(object, objectGenerator, null,
+                    isDeduplicateObjects(object.getClass()) ? new JsonPointerTracker(null, "/") : null);
+        }
         return objectGenerator.getResult();
     }
 
+    private List objectToList(Object object) {
```
rmannibucau: We already have that kind of logic, therefore question is: cant we reuse it and avoid one array and list allocation OR at least reuse the existing code?

elexx:
I couldn't find another way to cast/convert an arbitrary `Object` to an array without copying it.

But of course, if there is already such a method in the code base reusing it is the way to go. I had a quick look, do you mean `org.apache.johnzon.jsonb.JohnzonJsonb#toArray`? If so, it would be in the wrong package to be used by `Mapper` - any suggestions where I shall move it to?

rmannibucau:
@elexx it is not a package issue I think but a responsability issue. Mapper should only delegate to mapper generator and mapper parser now so the method belongs to MappingGeneratorImpl. If you look writeArray it betters handle that logic IMHO and I don't see any blocker to enable to use that logic from the mapper. Wdyt?

elexx:
You are right! Please have a look at the commit I just pushed. - Maybe I should squash them to one commit, before this PR gets merged to master?
 
Quick question, why does `toStructure` handle Primitives on it's own, instead of delegating to the Generator? Shouldn't `MappingGeneratorImpl#doWriteObject` and `#writePrimitives` take care of them?

rmannibucau:
@elexx this looks good to me now, thanks a lot. About toStructure I guess it is cause I had in mind that the generator was more linked to JsonGenerator so arrays and objects but I tend to agree it should move there as well.

@struberg want to have a look before we merge?

struberg:
right now really busy with other stuff. But from the discussion and a quick look over the code it looks fine. Go on!

PR #39, opened 2019-04-03
Deserialization of primitives

kalgon:
Fixed deserialization of primitives.

Serializers, Deserializers and Adapters ParameterizedTypes can be retrieved from super-classes/interfaces.
I created a utility class Types for that as I did not want to add commons-lang or any other libs.

Feedback is welcome!

### DIFF
--- a/johnzon-core/src/main/java/org/apache/johnzon/core/Types.java
+++ b/johnzon-core/src/main/java/org/apache/johnzon/core/Types.java
@@ -17,105 +17,68 @@
  * under the License.
  */
 package org.apache.johnzon.core;
-
-import java.lang.reflect.GenericArrayType;
+    
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
 import java.lang.reflect.TypeVariable;
-import java.lang.reflect.WildcardType;
 import java.util.Arrays;
 
 public class Types {
-
-    public static class TypeVisitor<T> {
-
-        public T visit(Class<?> type) {
-            throw new UnsupportedOperationException("Visiting Class not supported.");
+    
+    private static Type[] resolveArgumentTypes(Class<?> type, Class<?> superType) {
+        if (superType.equals(type)) {
+            // May return Class[] instead of Type[], so copy it as a Type[] to avoid
+            // problems in visit(ParameterizedType)
+            return Arrays.copyOf(type.getTypeParameters(), superType.getTypeParameters().length, Type[].class);
         }
-
-        public T visit(GenericArrayType type) {
-            throw new UnsupportedOperationException("Visiting GenericArrayType not supported.");
+        if (type.getSuperclass() != null && superType.isAssignableFrom(type.getSuperclass())) {
+            return resolveArgumentTypes(type.getGenericSuperclass(), superType);
         }
-
-        public T visit(ParameterizedType type) {
-            throw new UnsupportedOperationException("Visiting ParameterizedType not supported.");
-        }
-
-        public T visit(TypeVariable<?> type) {
-            throw new UnsupportedOperationException("Visiting TypeVariable not supported.");
-        }
-
-        public T visit(WildcardType type) {
-            throw new UnsupportedOperationException("Visiting WildcardType not supported.");
-        }
-
-        public final T visit(Type type) {
-            if (type instanceof Class<?>) {
-                return visit((Class<?>) type);
+        Class<?>[] interfaces = type.getInterfaces();
+        Type[] genericInterfaces = type.getGenericInterfaces();
+        for (int i = 0; i < interfaces.length; i++) {
+            if (superType.isAssignableFrom(interfaces[i])) {
+                return resolveArgumentTypes(genericInterfaces[i], superType);
             }
-            if (type instanceof ParameterizedType) {
-                return visit((ParameterizedType) type);
-            }
-            if (type instanceof WildcardType) {
-                return visit((WildcardType) type);
-            }
-            if (type instanceof TypeVariable<?>) {
-                return visit((TypeVariable<?>) type);
-            }
-            if (type instanceof GenericArrayType) {
-                return visit((GenericArrayType) type);
-            }
-            throw new IllegalArgumentException(String.format("Unknown type: %s", type.getClass()));
         }
+        throw new IllegalArgumentException(String.format("%s is not assignable from %s", type, superType));
     }
-
-    private static class ArgumentTypeResolver extends TypeVisitor<Type[]> {
-
-        private final Class<?> superType;
-
-        public ArgumentTypeResolver(Class<?> superType) {
-            this.superType = superType;
-        }
-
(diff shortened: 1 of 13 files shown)


johnzon-mapper/src/test/java/org/apache/johnzon/mapper/ObjectTypeTest.java
```diff
@@ -205,8 +207,8 @@ public void writeJson(Dog instance, MappingGenerator mappingGenerator) {
         }
 
         @Override
-        public Dog fromJson(JsonObject jsonObject, Type targetType, MappingParser parser) {
-            String javaType = jsonObject.getString("//javaType");
+        public Dog fromJson(JsonValue jsonObject, Type targetType, MappingParser parser) {
+            String javaType = ((JsonObject) jsonObject).getString("//javaType");
```
rmannibucau: asJsonObject() ;)

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonValueParserAdapter.java
```diff
@@ -0,0 +1,146 @@
...
+
+        @Override
+        public BigDecimal getBigDecimal() {
+            return getValue().bigDecimalValue();
+        }
+    }
+    
+    public static JsonParser createFor(JsonValue jsonValue, Supplier<JsonParserFactory> parserFactoryProvider) {
+        if (jsonValue instanceof JsonObject) {
+            return parserFactoryProvider.get().createParser((JsonObject) jsonValue);
```
rmannibucau: we can use asJsonXXX flavors for objects/arrays

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonValueParserAdapter.java
```diff
@@ -0,0 +1,146 @@
...
+        }
+
+        @Override
+        public BigDecimal getBigDecimal() {
+            return getValue().bigDecimalValue();
+        }
+    }
+    
+    public static JsonParser createFor(JsonValue jsonValue, Supplier<JsonParserFactory> parserFactoryProvider) {
+        if (jsonValue instanceof JsonObject) {
```
rmannibucau: switch on getValueType()?

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonValueParserAdapter.java
```diff
@@ -0,0 +1,146 @@
...
+        this.jsonValue = jsonValue;
+    }
+
+    @Override
+    public boolean hasNext() {
+        return false;
+    }
+
+    @Override
+    public Event next() {
```
rmannibucau: makes a lot of unsupported, wonder if the same trick than on the write side can be applied, i.e wrapping the actual impl and just having a small event buffer to handle primitive/object cases, wdyt? did you estimate it?

johnzon-core/src/main/java/org/apache/johnzon/core/Types.java
```diff
@@ -0,0 +1,157 @@
...
+package org.apache.johnzon.core;
+
+import java.lang.reflect.GenericArrayType;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+import java.lang.reflect.TypeVariable;
+import java.lang.reflect.WildcardType;
+import java.util.Arrays;
+
+public class Types {
```
rmannibucau: this is a bit complex compared to preview impl so can be worth a comment on why it was refactored

johnzon-core/src/main/java/org/apache/johnzon/core/Types.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public static class TypeVisitor<T> {
```
rmannibucau: at the end we only support ParameterizedType and Class - for now - right? So why do we need so much hooks - I'm thinking to just a visitDefault to be concrete? wdyt?

kalgon:
Well, I first took a look at what you did for serialization (DynamicMappingGenerator) but was unable to do the same for deserialization. I had to change the signature of ObjectConverter.Reader because the only places where I could find Deserializers are JohnzonBuilder and JsonbAccessMode (where they are converted into ObjectConverter.Readers which need a JsonObject and not a JsonValue).

About the serialization, I added some more tests (in MoreTests) which are still failing if you uncomment some the annotations. I was wondering, if you have a json property of type X[] or List<X> and you specify a JsonbSerializer on that property, on what type does the serializer act? on X or on X[]/List<X> ?

Concerning the ParameterizedType, on one of my projects, I had some abstract Jsonb(De)Serializers and johnzon was complaining that they did not implement the correct interfaces (because it was only looking for the PT on the concrete class only).

rmannibucau:
@kalgon for arrays/lists we match the type for the serializer, if it matches the container we apply it globally otherwise we must apply it by item IIRC.
About Types, I'm not against it at all but we should probably ensure it is 1. like Mapping an instance per mapper to have customizations later and not static, 2. ensure we write the hierarchical case, I have to admit I understood only cause I saw worse in openwebbeans but it is not obvious looking from the johnzon window so just a small comment (A <| B and B implements X etc) would solve that point.
I'm happy to apply it - pby tomorrow through - if you can fix the Types point and just leave a note on the tests you need help, i will try to have a look.

kalgon:
I tried to clarify what Types is used for. The failing test can be found in MoreTests, just uncomment the @JsonbTypeSerializer/@JsonbTypeDeserializer annotations inthe Wrapper class. You will see that the serializers are not used while the deserializers are.

johnzon-core/src/main/java/org/apache/johnzon/core/Types.java
```diff
@@ -0,0 +1,125 @@
...
+        public Type getOwnerType() {
+            return null;
+        }
+
+        @Override
+        public Type[] getActualTypeArguments() {
+            return arguments;
+        }
+
+        // TODO equals, hashcode, toString if needed
```
rmannibucau: side note: not mandatory today but if desired you can steal these methods from https://github.com/apache/johnzon/blob/master/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/reflection/JohnzonParameterizedType.java (don't link the two modules since mapper should run with another json-p impl IIRC)

rmannibucau:
/me is fetching the PR and checking the MoreTests - we can need to rename it BTW ;)

rmannibucau:
@kalgon , hey Xavier, applied your patch with a few changes, do you care having a look please? (FYI i renamed MoreTests "SeriaizersRoundTripTest" to find it more easily ;))

kalgon:
Hi Romain, I took a look at your patch and I have nothing to add! Great work! Thanks a lot!

PR #44, opened 2019-09-18
Fix for [JOHNZON-276]: StackOverflowError

mkarg:
The following code produces a StackOverflowError:
```java
@JsonVisibility(DefaultPropertyVisibilityStrategy.class)
public final class C
{
   @JsonProperty private boolean foo;
}
```
The problem is caused by the fact that this strategy tries to compare "this" by "==" with a new instance of itself, which certainly fails, because the new copy is definitively a different reference than the one currently executed and instantiated due to the class annotation.

The fix is simple: Adding an equals method just comparing the class type (as the strategy has no members other than a cache), and calling that instead of "==".

### DIFF
--- a/johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/DefaultPropertyVisibilityStrategy.java
+++ b/johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/DefaultPropertyVisibilityStrategy.java
@@ -40,14 +40,14 @@ public class DefaultPropertyVisibilityStrategy implements javax.json.bind.config
         }
         final PropertyVisibilityStrategy strategy = strategies.computeIfAbsent(
                 field.getDeclaringClass(), this::visibilityStrategy);
-        return strategy == this ? Modifier.isPublic(field.getModifiers()) : strategy.isVisible(field);
+        return this.equals(strategy) ? Modifier.isPublic(field.getModifiers()) : strategy.isVisible(field);
     }
 
     @Override
     public boolean isVisible(final Method method) {
         final PropertyVisibilityStrategy strategy = strategies.computeIfAbsent(
                 method.getDeclaringClass(), this::visibilityStrategy);
-        return strategy == this ? Modifier.isPublic(method.getModifiers()) : strategy.isVisible(method);
+        return this.equals(strategy) ? Modifier.isPublic(method.getModifiers()) : strategy.isVisible(method);
     }
 
     private PropertyVisibilityStrategy visibilityStrategy(final Class<?> type) { // can be cached
@@ -86,5 +86,10 @@ public class DefaultPropertyVisibilityStrategy implements javax.json.bind.config
             }
         }
         return this;
+    }
+
+    @Override
+    public boolean equals(final Object o) {
+        return o instanceof DefaultPropertyVisibilityStrategy;
     }
 }


rmannibucau:
Hi @mkarg ,

There is one thing worrying me in that case - and this is why "==" was intentinoally used: DefaultPropertyVisibilityStrategy is an internal class - not sure it should have been public - but it is not intended to be used by jsonb users. So in practise you can't be in the "equals" case IMHO.

Romain

mkarg:
@rmannibucau That might be true, but does Johnzon really want to write code that *has potential* to fail? ;-) Whether or not this was intentional - I *did* run into that problem yesterday because the class *is public*, and I *had* to use it to overcome another bug of Johnzon: Leaving out the explicit `@JsonbVisibility` does *not* invoke the field writer (while it detects the `@JsonbProperty`, it still only invokes the field writer when the field is `public`)! So either we need *this* PR, or we need a fix for that original bug (which I cannot deliver on short term due to missing time).

rmannibucau:
@mkarg no issue on the time point I really understand, my question was really: dont we just want to make it "private" (I know we cant do "private class" but we can hide it from user land).

cc @struberg wdyt? should we expose that class? Wdyt?

mkarg:
@rmannibucau I'm ok with hiding it from userland, but despite that, the "reference ==" idiom is really error prone and brings hardly no measureable performance benefit compared to the "safe and clean" solution proposed by this PR. So even if we remove "public" we should fix that. I spent a full day because of these two bugs... ;-(

rmannibucau:
Btw equals as instanceof does not work, think about subclassing for example
I'm personally keen to just hide it since its "api" is not intended to be stable

Do you need such a default impl or was the explicit def just a test?

mkarg:
I don't see why you mention subclassing, as the current `this ==` clearly is unable to deal with subclassing *already*. So you reject my code for something that the original code already fails with?

Anyways, as I wrote above I had to _explicitly_ refer to this strategy using `@JsonVisibility` on a POJO to work around *another* bug of Johnzon:
```java
@JsonVisibility(DefaultPropertyVisibilityStrategy.class)
public final class C
{
   @JsonProperty private boolean foo;

   public final boolean isFoo() { return foo; }
}
```
*Without* `@JsonVisibility` the POJO's field `foo` is **never** written on JSON parsing but stays `false` forever. Making it `public` would work around that, but is not intended. I *want* to keep it `private` as it shall be immutable by Java code, while JSON-B *still shall* modify it.

I debugged this, and noticed that Johnzon actually knows that it *wants* to use the getter method + the field writer as accessors, but actually just doesn't invoke the field writer when it comes to parsing JSON. Had no time to debug it deeper, that's why I didn't report *that* first.

rmannibucau:
@mkarg == enable to subclass the method and keep it working/handlable (I agree it requires a threadlocal which is ugly but once again this class must never be used in user code), instanceof breaks that directly bypassing subclass logic all the time. Also note I don't reject the change but want to highlight it would require some change and is not needed since your use case works on master (1.2.0 I guess) without any visibility hack.

mkarg:
@rmannibucau Actually the use case FAILS on master (1.2.0-SNAPSHOT) and NEEDS the visibility hack! Regarding subclasses, we could replace `instanceof` by `Class.isAssignableFrom`. Still no need to use `==` or a threadlocal hack. BTW, AFAIK `o instanceof X` will be true even if o is an instance of a *subclass of X*, so it *will* work with subclasses even with `instanceof`.

rmannibucau:
@mkarg isAssignableFrom is the same as instanceof, it still ignores the subclassing we could use later on internally. The fact instanceof is true for subclass is what makes it not working, it bypass the subclass impl instead of delegating to it, was my point. Can you review the following test which describe both the inheritance and case you try to achieve which works: https://github.com/apache/johnzon/blob/master/johnzon-jsonb/src/test/java/org/apache/johnzon/jsonb/DefaultPropertyVisibilityStrategyTest.java#L33 ?

mkarg:
@rmannibucau Thanks for the test case! But strange, the test succeeds pretty well while the same code fails with the reported bugs. It seems I *do* have to debug it in the end. Possibly a third bug not covered by the test case so far... ;-(

rmannibucau:
@mkarg if you can push a github repo with a failling test i can try to have a look

mkarg:
@rmannibucau Thanks a lot for this kind offer! Today I have to fix some other stuff before I have time to look into this particular problem again. If I need your help, I am pretty sure I will pick up your offer then! :-)

mkarg:
> @mkarg if you can push a github repo with a failling test i can try to have a look

@rmannibucau Again, thanks for this kind offer. In fact, just now I resolved the case on my own (by using latest Johnzon 1.2.1 including all the fixes of the last weeks), plus write an adapter. So I now was able to get rid of the explicit `@JsonVisibility(DefaultPropertyVisibilityStrategy.class)`.

Having said that, I *now* do agree to your original claim: I do **not** see a need for `DefaultPropertyVisibilityStrategy` to be `public` anymore now that 1.2.1 is published. :-)

So the question is, how to go on with this PR / Issue 276?

rmannibucau:
We can simply close it I think.
What's your opinion?

mkarg:
I think as soon as the class is non-public, we can close it.

rmannibucau:
@mkarg do you do it or do you let us do (both ways are fine for me, just don't want to eat your work ;))

mkarg:
> @mkarg do you do it or do you let us do (both ways are fine for me, just don't want to eat your work ;))

Apparently we share the same sense of humor. ;-)

I will do it. Actually I already did it, just need a few seconds to do the needed JIRA + Github book keeping... Stay tuned.

mkarg:
@rmannibucau Here it is: https://github.com/apache/johnzon/pull/54 :-)

mkarg:
According to general agreement among the contributors, this issue will not get fixed.

PR #47, opened 2019-09-22
Enabled test as it seems to run well.

mkarg:
Tried to reproduce the bug multiple times on Windows and Linux, but it did not happen anymore. The reason why this test was disabled seems to be gone. Hence it seems it was a local instability on my side, and it should be safe to enable this test now. Sorry for the confusion! :-)


mkarg:
@rmannibucau Review please. Sorry for the confusion! There is no bug apparently... :-)

PR #50, opened 2019-09-24
JOHNZON-272 Incorrect EOL created on Windows

mkarg:
On Windows, `.java` source files do have `CR LF` endings typically. If creating `LF` endings only (just like on Linux), native Windows editors will show the resulting source file as one single line. To work correctly on **all** operating systems, this change does _not_ imply _any_ particular EOL sequence, but offloads this task to the JRE.

### DIFF
--- a/johnzon-maven-plugin/src/main/java/org/apache/johnzon/maven/plugin/ExampleToModelMojo.java
+++ b/johnzon-maven-plugin/src/main/java/org/apache/johnzon/maven/plugin/ExampleToModelMojo.java
@@ -35,11 +35,10 @@ import javax.json.JsonStructure;
 import javax.json.JsonValue;
 import java.io.File;
 import java.io.FileReader;
-import java.io.FileWriter;
+import java.io.PrintWriter;
 import java.io.FilenameFilter;
 import java.io.IOException;
 import java.io.StringWriter;
-import java.io.Writer;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.Iterator;
@@ -99,7 +98,7 @@ public class ExampleToModelMojo extends AbstractMojo {
     }
 
     // TODO: unicity of field name, better nested array/object handling
-    private void generate(final JsonReaderFactory readerFactory, final File source, final Writer writer, final String javaName) throws MojoExecutionException {
+    private void generate(final JsonReaderFactory readerFactory, final File source, final PrintWriter writer, final String javaName) throws MojoExecutionException {
         JsonReader reader = null;
         try {
             reader = readerFactory.createReader(new FileReader(source));
@@ -113,25 +112,25 @@ public class ExampleToModelMojo extends AbstractMojo {
 
             // while we browse the example tree just store imports as well, avoids a 2 passes processing duplicating imports logic
             final StringWriter memBuffer = new StringWriter();
-            generateFieldsAndMethods(memBuffer, object, "    ", imports);
+            generateFieldsAndMethods(new PrintWriter(memBuffer), object, "    ", imports);
 
             if (header != null) {
-                writer.write(header);
-                writer.write('\n');
+                writer.printf(header).println();
             }
 
-            writer.write("package " + packageBase + ";\n\n");
+            writer.println("package " + packageBase + ";");
+            writer.println();
 
             if (!imports.isEmpty()) {
                 for (final String imp : imports) {
-                    writer.write("import " + imp + ";\n");
+                    writer.println("import " + imp + ";");
                 }
-                writer.write('\n');
+                writer.println();
             }
 
-            writer.write("public class " + javaName + " {\n");
+            writer.println("public class " + javaName + " {");
             writer.write(memBuffer.toString());
-            writer.write("}\n");
+            writer.println("}");
         } catch (final IOException e) {
             throw new MojoExecutionException(e.getMessage(), e);
         } finally {
@@ -141,7 +140,7 @@ public class ExampleToModelMojo extends AbstractMojo {
         }
     }
 
-    private void generateFieldsAndMethods(final Writer writer, final JsonObject object, final String prefix,
+    private void generateFieldsAndMethods(final PrintWriter writer, final JsonObject object, final String prefix,
                                           final Collection<String> imports) throws IOException {
         final Map<String, JsonObject> nestedTypes = new TreeMap<String, JsonObject>();
         {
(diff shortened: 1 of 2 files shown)


rmannibucau:
Hi @mkarg, the intention was to only use \n for source files - as most java projects - and not let the generation be OS dependent. This works well on windows (except using notepad.exe but it is fine for the determistic generation gain which is way more important than working in notepad.exe)

mkarg:
@rmannibucau The created files are shown as single-liners in native Windows editors, which is the opposit of "working well". So the intention of rather all "real" Windows guys is that the created file have CR LF endings, just as all .java source on Windows has.

rmannibucau:
@mkarg this is just not true, only a few editors - with other issues BTW, like the lack of UTF-8 support, do it (ie notepad.exe), most other editors work well (tested notepad++, atom, vscode, netbeans, idea, eclipse, geany to cite a few). Finally note that these sources are "generated-sources" intended to be compiled and included (in -sources and main artifacts) and not to be edited so the predictability is more important than being OS dependent in any case (which means my config proposal maybe does not even make sense).

mkarg:
@rmannibucau The official sole EOL on Windows is CR LF, check Wikipedia. The editors you mention are not Windows-natives but simply migrations from other OS to Windows. NOTEPAD is not an exception, it is the _prototype_ of a "real" Windows editor. I understand that these sources are just generated, but still they should be human-readable in "real" Windows editors on Windows, not just in migrations from other OS or cross-platform editors. As a side effect, no tricks and no doc changes are needed on Windows. The world is more than Linux. ;-)

rmannibucau:
@mkarg I know all that but it does not change the fact that as an user of that plugin I expect the generated files to be the same on any OS to produce the same artifacts on any OS and not be developer dependent as much as possible. This is the rough minimal requirement for this generator. I can hear you disagree with that but then the solution is a toggle to merge both requirements with a default to LF for backward compatibility.

mkarg:
@rmannibucau Let's agree to disagree. ;-) Anyways, I am not the project lead. Either you are (I actually do not know - where can I lookup that information?) so you have to decide on your own, or we have to find another "real" Windows user and ask him to get a majority. :-)

cyberphone:
Notepad has after a 20 year standstill finally been upgraded to understand that \n is the industry standard :)

I personally use Windows as my main development OS and have turned off EOL normalization in git.  Previously I used Notepad to verify that it was working but with the upgrade that doesn't work anymore...

mkarg:
@cyberphone I would be careful with "being _the_ industry standard". Each OS works differently, and in the past 30+ years I have been in this industry, I have seen so many things people called "industry standard" come and go, that I learned to simply respect that each technology / OS has its own rules and follow them "as-is", which might be inconvenient and not my personal choice, but always leads to being a "good citizen".

cyberphone:
@mkarg Let me put it like this,  all tool vendors of any significance can now deal with text files using \n as EOL mark.  Microsoft though still struggles with UTF-8 where VS-2017 by default inserts a BOM which causes several non-Microsoft tools to crash including Jenkins for `gh-pages`.

rmannibucau:
Even if my past experience tend to join @cyberphone on the state (my first years of work were on windows and standard - where I was working - was already \n for any java source file and I never worked on a single project not using that since), I agree with @mkarg it is a choice for its own code. Now here we generate sources for build and source inclusion (for maven auto download sources feature of _IDE_), not for editing so should we make it configurable or not is the only question IMHO and not convinced it is that critical on my side but happy to add a config param if it helps anyone (and it is a small detail of the project, we don't need to take that much time ;)).

mkarg:
@cyberphone Still you only look on the Linux/Windows universe, but there are lot of other OS that existed in the past and might exist in future. PR #50 will work with *any* EOL. ;-)

mkarg:
@rmannibucau Maybe I was misunderstood. I have nothing against you adding this ontop of PR #50. I just won't do that on my own as I do not see the need. ;-)

rmannibucau:
@mkarg well, #50 is not mergeable like that since it is a regression for all consumers I know (ok not a ton but concrete feedback), this is my main point, so can't be merged without being reworked in current state :(.

cyberphone:
@mkarg Right, the Unicode consortium have defined a Line Separator character:
https://stackoverflow.com/questions/3072152/what-is-unicode-character-2028-ls-line-separator-used-for
I think they have had some "minor" difficulties getting it adopted :)

I let you guys decide in peace, while I continue decorating my `.gitattributes` files with `* -text`.

mkarg:
@rmannibucau Then we have to keep this issue open until doomsday, as for me it is a bug fix as producing LF-only EOLs on Windows is definitively a bug (yes, I understand you want it that way). ;-)

mkarg:
@rmannibucau Maybe a working concensus would be that you send me a PR against my fork so your commit sits ontop? As a result this PR #50 automatically would contain your "fix". I could live with that, and we both get what we want in the end.

rmannibucau:
@mkarg well, as mentionned I more keen - personally - to not do anything in terms of code and just document that git must be configured to not corrupt the original sources. My absolute will is to prevent regressions so if you want an OS dependent option, I'm fine with it, but it must be a toggle we can enable with a default and not have it activated OOTB.

mkarg:
@rmannibucau You call it a regression, I call it the sole correct fix for a severe bug. We simply have completely diametrical views on this topic (and both are equally valid from within each other's universe). As it seems we cannot find a concensus on this topic, so it makes no sense that we spend any more time into this. I'm good with keeping Johnzon being buggy in this regard and step forward to fix other issues. BTW, I still do not see what you want to gain with the doc change. The `master` branch currently *passes* all tests on Windows *with* git performing its default strategy on Windows (= add CR). So if you change the docs, you simply make things worse, as then the test will *fail* again.

rmannibucau:
@mkarg test will not test with the doc change since it builds well on linux as well - wouldn't have been merged otherwise ;). The advantage to update the doc is nobody can encounter the case you got which is not a case we take into account because we (wrongly it seems ;)) assumed all the universe was using the setup Anders mentionned. Now no issue to keep it like that too since we didn't get windows user complains before even if some built/contributed to the project already (guess git was correctly configured to use the common \n as EOL). Alternatively we can add a .gitattributes (like https://github.com/Talend/component-runtime/blob/288b4b2854e59f6d1de7128f188fddd514e9d65d/.gitattributes) which would have prevented your issue to happen. Can be simpler after all. Wdyt?

mkarg:
@rmannibucau Sigh - Agreed after reading the Java Langage Specification (for those who care: It says that Java source files MUST end on LF or CR LF or CR, which explicitly means that EOLs other than that MUST NOT be used, which means my code in PR #50 will fail on those OS's having other line separators) -- So I MUST close this PR as invalid. As apparently it is Johnzon's explicit intention to have that particular file being untouched and encoded explicitly with LF, there must not be a *general* doc change nor a *general* .gitattribute, but just a .gitattribute for this single file in the Johnzon repo -- all other files must stay under the control of the local OS's / local user's decision. But as 45408012b5ded708134d6e43ae587d0645b64f37 actually fixes the same problem, I cannot see the benefit *unless* you undo that commit. So is *replacing that commit* what you prose? Just to make me understand your actual intention.

mkarg:
@rmannibucau  I have authored a PR with the solution outlined by you: #51. :-)

PR #51, opened 2019-09-24
JOHNZON-272 Fails to build Maven Plugin - Alternative Solution

mkarg:
Using `.gitattributes` to guarantee LF line endings instead of explicitly replacing wrong line endings in Java code.

See https://github.com/apache/johnzon/pull/50#issuecomment-534668791.

I confirm it actually solves JOHNZON-272 on both, Linux (tried with Gitpod) and Windows (tried with Windows 8.1).


mkarg:
@rmannibucau Review please. And thanks for all your patient guidance. :-)

PR #52, opened 2019-09-26
JOHNZON-281 JAX-RS Provider MUST throw NoContentException

mkarg:
The Johnzon JAX-RS Provider throws the wrong exception type in case of reading empty input streams. JAX-RS 2.1 requires to throw **NoContentException**, but Johnzon throws **JsonbException**.

See JavaDocs of JAX-RS 2.1 MessageBodyReader:

_"In case the entity input stream is empty, the reader is expected to either return a Java representation of a zero-length entity or throw a javax.ws.rs.core.NoContentException in case no zero-length entity representation is defined for the supported Java type."_

### DIFF
--- a/johnzon-jsonb/src/main/java/org/apache/johnzon/jaxrs/jsonb/jaxrs/JsonbJaxrsProvider.java
+++ b/johnzon-jsonb/src/main/java/org/apache/johnzon/jaxrs/jsonb/jaxrs/JsonbJaxrsProvider.java
@@ -29,6 +29,7 @@ import javax.ws.rs.Produces;
 import javax.ws.rs.WebApplicationException;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.core.NoContentException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.core.StreamingOutput;
 import javax.ws.rs.ext.MessageBodyReader;
@@ -37,6 +38,7 @@ import javax.ws.rs.ext.Provider;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
+import java.io.PushbackInputStream;
 import java.io.Reader;
 import java.io.StringReader;
 import java.io.Writer;
@@ -187,7 +189,23 @@ public class JsonbJaxrsProvider<T> implements MessageBodyWriter<T>, MessageBodyR
 
     @Override
     public T readFrom(final Class<T> type, final Type genericType, final Annotation[] annotations, final MediaType mediaType,
-            final MultivaluedMap<String, String> httpHeaders, final InputStream entityStream) throws WebApplicationException {
+            final MultivaluedMap<String, String> httpHeaders, InputStream entityStream) throws IOException, WebApplicationException {
+        if (entityStream.markSupported()) {
+            entityStream.mark(1);
+            if (entityStream.read() == -1) {
+                throw new NoContentException("Johnzon cannot parse empty input stream");
+            }
+            entityStream.reset();
+        } else {
+            final PushbackInputStream buffer = new PushbackInputStream(entityStream);
+            int firstByte = buffer.read();
+            if (firstByte == -1) {
+                throw new NoContentException("Johnzon cannot parse empty input stream");
+            }
+            buffer.unread(firstByte);
+            entityStream = buffer;
+        }
+
         return getJsonb(type).fromJson(entityStream, genericType);
     }
 
--- a/johnzon-jsonb/src/main/java/org/apache/johnzon/jaxrs/jsonb/jaxrs/JsonbJaxrsProvider.java
+++ b/johnzon-jsonb/src/main/java/org/apache/johnzon/jaxrs/jsonb/jaxrs/JsonbJaxrsProvider.java
@@ -38,7 +38,6 @@ import javax.ws.rs.ext.Provider;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
-import java.io.PushbackInputStream;
 import java.io.Reader;
 import java.io.StringReader;
 import java.io.Writer;
@@ -55,7 +54,13 @@ import javax.ws.rs.core.Context;
 import javax.ws.rs.ext.ContextResolver;
 import javax.ws.rs.ext.Providers;
 
+import org.apache.johnzon.core.JsonReaderImpl.NothingToRead;
+
 // here while we dont compile in java 8 jaxrs module, when migrated we'll merge it with IgnorableTypes hierarchy at least
+/**
+ * JAX-RS Compliance Warning: This MBR is only compliant with JAX-RS 2.1 as long as it runs on Johnzon JSON-P. Using other
+ * JSON-P implementations will result in <em>not</em> throwing {@link NoContentException} for empty input streams.
+ */
 @Provider
 @Produces(MediaType.APPLICATION_JSON)
 @Consumes(MediaType.APPLICATION_JSON)
@@ -80,6 +85,11 @@ public class JsonbJaxrsProvider<T> implements MessageBodyWriter<T>, MessageBodyR
 
     private boolean isIgnored(final Class<?> type) {
         return ignores != null && ignores.contains(type.getName());
+    }
+
+    // simplifies testing without mocking content injection
+    void setProviders(final Providers providers) {
+        this.providers = providers;
     }
 
     // config - main containers support the configuration of providers this way
@@ -190,23 +200,11 @@ public class JsonbJaxrsProvider<T> implements MessageBodyWriter<T>, MessageBodyR
     @Override
     public T readFrom(final Class<T> type, final Type genericType, final Annotation[] annotations, final MediaType mediaType,
             final MultivaluedMap<String, String> httpHeaders, InputStream entityStream) throws IOException, WebApplicationException {
-        if (entityStream.markSupported()) {
-            entityStream.mark(1);
-            if (entityStream.read() == -1) {
-                throw new NoContentException("Johnzon cannot parse empty input stream");
-            }
-            entityStream.reset();
-        } else {
-            final PushbackInputStream buffer = new PushbackInputStream(entityStream);
-            int firstByte = buffer.read();
-            if (firstByte == -1) {
-                throw new NoContentException("Johnzon cannot parse empty input stream");
-            }
-            buffer.unread(firstByte);
-            entityStream = buffer;
+        try {
+            return getJsonb(type).fromJson(entityStream, genericType);
+        } catch (final NothingToRead e) {
+            throw new NoContentException(e);
         }
-
-        return getJsonb(type).fromJson(entityStream, genericType);
     }
 
     @Override
--- a/johnzon-jaxrs/src/main/java/org/apache/johnzon/jaxrs/JohnzonMessageBodyReader.java
+++ b/johnzon-jaxrs/src/main/java/org/apache/johnzon/jaxrs/JohnzonMessageBodyReader.java
@@ -18,6 +18,7 @@
  */
 package org.apache.johnzon.jaxrs;
 
+import org.apache.johnzon.core.JsonReaderImpl.NothingToRead;
 import org.apache.johnzon.mapper.Mapper;
 import org.apache.johnzon.mapper.MapperBuilder;
 
@@ -30,7 +31,6 @@ import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.MessageBodyReader;
 import java.io.IOException;
 import java.io.InputStream;
-import java.io.PushbackInputStream;
 import java.io.Reader;
 import java.lang.annotation.Annotation;
 import java.lang.reflect.ParameterizedType;
@@ -38,6 +38,10 @@ import java.lang.reflect.Type;
 import java.util.Collection;
 
 // @Provider // don't let it be scanned, it would conflict with JohnzonProvider
+/**
+ * JAX-RS Compliance Warning: This MBR is only compliant with JAX-RS 2.1 as long as it runs on Johnzon JSON-P. Using other
+ * JSON-P implementations will result in <em>not</em> throwing {@link NoContentException} for empty input streams.
+ */
 @Consumes({
     "application/json", "*/json",
     "*/*+json", "*/x-json",
@@ -71,27 +75,15 @@ public class JohnzonMessageBodyReader<T> extends IgnorableTypes implements Messa
                       final Annotation[] annotations, final MediaType mediaType,
                       final MultivaluedMap<String, String> httpHeaders,
                       InputStream entityStream) throws IOException {
-        if (entityStream.markSupported()) {
-            entityStream.mark(1);
-            if (entityStream.read() == -1) {
-                throw new NoContentException("Johnzon cannot parse empty input stream");
+        try {
+            if (rawType.isArray()) {
+                return (T) mapper.readArray(entityStream, rawType.getComponentType());
+            } else if (Collection.class.isAssignableFrom(rawType) && ParameterizedType.class.isInstance(genericType)) {
+                return (T) mapper.readCollection(entityStream, ParameterizedType.class.cast(genericType));
             }
-            entityStream.reset();
-        } else {
-            final PushbackInputStream buffer = new PushbackInputStream(entityStream);
-            int firstByte = buffer.read();
-            if (firstByte == -1) {
-                throw new NoContentException("Johnzon cannot parse empty input stream");
-            }
-            buffer.unread(firstByte);
-            entityStream = buffer;
+            return mapper.readObject(entityStream, genericType);
+        } catch (final NothingToRead e) {
+            throw new NoContentException(e);
         }
-
-        if (rawType.isArray()) {
-            return (T) mapper.readArray(entityStream, rawType.getComponentType());
-        } else if (Collection.class.isAssignableFrom(rawType) && ParameterizedType.class.isInstance(genericType)) {
-            return (T) mapper.readCollection(entityStream, ParameterizedType.class.cast(genericType));
-        }
-        return mapper.readObject(entityStream, genericType);
     }
 }
--- a/johnzon-jaxrs/src/main/java/org/apache/johnzon/jaxrs/JohnzonMessageBodyReader.java
+++ b/johnzon-jaxrs/src/main/java/org/apache/johnzon/jaxrs/JohnzonMessageBodyReader.java
@@ -25,10 +25,12 @@ import javax.json.JsonStructure;
 import javax.ws.rs.Consumes;
 import javax.ws.rs.core.MediaType;
 import javax.ws.rs.core.MultivaluedMap;
+import javax.ws.rs.core.NoContentException;
 import javax.ws.rs.core.Response;
 import javax.ws.rs.ext.MessageBodyReader;
 import java.io.IOException;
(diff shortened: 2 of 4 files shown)


mkarg:
This PR proposes a solution which reconstitutes compliance with JAX-RS 2.1 with a two-step approach to detect an empty input stream:

* First it tries to use mark/reset as this is very fast, but not guaranteed to be supported by all kinds of input streams.
* If this is not possible, then it will actively buffer the first byte, which is rather slow, but will work with all kinds of input stream.

johnzon-jsonb/src/main/java/org/apache/johnzon/jaxrs/jsonb/jaxrs/JsonbJaxrsProvider.java
```diff
@@ -187,7 +189,23 @@ public long getSize(final T t, final Class<?> type, final Type genericType, fina
 
     @Override
     public T readFrom(final Class<T> type, final Type genericType, final Annotation[] annotations, final MediaType mediaType,
-            final MultivaluedMap<String, String> httpHeaders, final InputStream entityStream) throws IOException, WebApplicationException {
+            final MultivaluedMap<String, String> httpHeaders, InputStream entityStream) throws IOException, WebApplicationException {
+        if (entityStream.markSupported()) {
```
rmannibucau: hey @mkarg, I would like to mention these few points before moving forward, please let me know what you think about it:

1. here the stream will rarely be markable (with cxf it will not most of the time) so we can surely skip this case
2. we already have a pushback inputstream for the jsonparser so i would consider adding a callback here we can register through the parser map. High level it would be a callback propagated through https://github.com/apache/johnzon/blob/master/johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JohnzonBuilder.java#L837 from Jsonb default instance to the parser and it would be called https://github.com/apache/johnzon/blob/156c7af77322fc0d11819756fc48fb18bd617b71/johnzon-core/src/main/java/org/apache/johnzon/core/RFC4627AwareInputStreamReader.java#L51 (this stream reader is created by the parser). The callback can then throw an exception the readFrom can catch and process in an exceptional way (see point 3)
3. trivial impl of this exceptional way can be throw new NoContentException but it also makes sense to have a way to respect the quote from the spec you did in the ticket and return an "empty" object IMHO, I assume it must be an option of the provider and handled in the catch of point 2 as well
4. you forget to commit the associated test? ;)

wdyt?

mkarg: @rmannibucau Thank you for the quick review. :-)
1. Actually I tried it out with the following result: Payara 4 really provides a markable stream, pure Jersey does not (or vice versa, I haven't actually written down), so dropping the mark case would _actually_ turn down performance at least for *some* configurations. So I would vote for keeping this case to provide *best possible* performance in each particular enviroment.
2. I will look into that, but quickly answered this sounds complex, error-prone and slow (try-catch IMHO is slower than double buffering). What would be the _actual_ benefit compared to my original proposal?
3. The intention of *this* PR is to provide JAX-RS 2.1 compliance **ASAP**. I see your point here and would love to have a more sophisticated solution, which e. g. checks if the requested class would allow creation of empty instances or not. But as the currently wrong code does not have this solution, and as it sounds non-trivial and slow, I think we must spend rather long time to get that attempt right _and_ fast _and_ bug-free. So IMHO we should do that in a _later_ PR _ontop_ of this one.
4. What is a test? ;-)

rmannibucau: 1. can be solved by a bench but from what you say mainstream is not markable (only payara?) so it shouldn't be the first citizen probably?
2. the benefit is the speed, overhead in O(1) vs O(size of the stream) (you can bench it on small, medium, huge streams)
3. this is not that hard or slow ;). My main point is we can't slow down a stream more than we do already, it is the slowest part we have and here we try to handle a case which never happens (literally) so any overhead in O(n) is negative for real users.
4. I didnt read that answer :D

mkarg: Thanks for the explanation.

In fact I do not see why the upfront markable check actually bothers you: It eats up rather zero time as the implementation in the usual suspects would be `return false;`? Anyways, I have no strong feeling about removing that case if you hate it so much. ;-)

Before I deep-dive into that code (I will do definitively): So what you claim is that `PushbackInputBuffer` actually is O(size of _stream_) -which would be a desaster!- instead of O(size of buffer) -which would be negligible as the buffer size is 1 byte-? Really? Didn't know that! Actually I should open a PR against the JRE to change that! ;-)

rmannibucau: The markable point is just in terms of maintenance but no other blocker so you can keep it, just ensure it works when false ;)

The pushback overhead is indeed o(n) for the same reason an arraylist if often replaced by an array in optimized impl (checkopened+length check), we dont need it here so it is just wasting resources to stack them.

N is the stream size, not the buffer one ;).

Hope it is clearer.

mkarg:
Apparently Johnzon has another MBR which I didn't know of before, and which also is incompliant to JAX-RS 2.1, so here are more commits:
* Fixed second MBR so now *both* MBRs are compliant to JAX-RS 2.1 (original claim of this PR)
* Added two unit tests to proof this claim (solves issue 4)

@rmannibucau I will definitively look into the open issues 2 and 3. This PR keeps "as wanted" issue 1 and solves issue 4, hence, the actual topic of this bug report is finished: Johnzon becomes JAX-RS 2.1 compliant again. Due to separation of concerns, I'd like to request that we merge PR #52 now, and I will open two new separate PRs *ontop*, one to address the performance regression (issue 2) and one to address the idea of returning *empty instances* (issue 3).

johnzon-jsonb/src/main/java/org/apache/johnzon/jaxrs/jsonb/jaxrs/JsonbJaxrsProvider.java
mkarg: I checked the code and you are right, this is something that we should address to prevent the O(n) overhead. I will look into that definitively, but to not further delay fixing of this bug (my application expects to catch `NoContentException` so I'm stuck without this) I'd kindly request that I may fix the performance regression *later this week*.

rmannibucau:
@mkarg I wouldnt fix the delegate MB impls, this spec point is not relevant in 100% of the cases so let's respect it only for the default provider (which is intended to ve sepc compliant compared to others which are intended to be composed and never auto-registered) with a low overhead and not for all other flavors

Side note: next release is coming, if you need some code help I can invest some time tomorrow or monday morning before the release launch

mkarg:
> @mkarg I wouldnt fix the delegate MB impls, this spec point is not relevant in 100% of the cases so let's respect it only for the default provider (which is intended to ve sepc compliant compared to others which are intended to be composed and never auto-registered) with a low overhead and not for all other flavors

In this case we should add clear warnings to those JavaDocs or even do not make them `public`. Which one is not intended for end-user application?

> Side note: next release is coming, if you need some code help I can invest some time tomorrow or monday morning before the release launch

It would be nice if you could write the performance code that you like to have as an optop commit.

rmannibucau:
@mkarg all are intended - and used - for end user applications but only the one you originally patched is intended to respect as much as possible JAX-RS.
This makes me think I have to remove the @Provider on the other one, I will take the point now cause it is a long overdue bug when the jar is scanned (it is not in apache products).
I will add the SPI I spoke about now and ping you back here.

mkarg:
@rmannibucau ok, I'm on stand-by

rmannibucau:
Hey @mkarg , just added a org.apache.johnzon.core.JsonReaderImpl$NothingToRead exception which should enable to do a try/catch (on IllegalStateException and then test the type with e.getClass().getName() to avoid to couple core module to jsonb probably).

mkarg:
@rmannibucau Thanks for that! Unfortunately I gotta go away for few hours (hopefully will continue later this night). In fact, before I finish the code, three questions:
* Once we catch the exception, the provider won't detect the zero-byte-stream when running ontop of other JSON-B implementations. Hence the provider is Johnzon-only. If this is intended, I will add a note in the JavaDocs.
* Shall we really compare exception message text? That's pretty ugly. I'd prefer to actually link the code with JSON-P and catch `NothingToRead` class.
* A MBR is still an MBR, hence a JAX-RS component, hence should comply with its rules. I could simply catch `NothingToRead` in that "other" MBR, too. This would be more clean than to say "this is not intended to be JAX-RS 2.1 compatible".

Wdyt?

rmannibucau:
Yes but this part of the jaxrs spec does not make sense at all so let the feature be specific (actually this jaxrs requirement breaks jsonb specified exceptions so can need to be reverted to let user use jsonb and add an exception mapper for this or just respect jsonb parsing and return a 400 which is more likely in most apps).

Not message but type, check out master ;). Thinking out loud, an exception mapper can make more sense and would be skipped when not loadable.

Agree a mbr is a mbr but jsonb-jaxrs integration is only in one provider, all other mbr dont have to comply to jaxrs (are not jsonb) and in 99% of the time they will be surfaced anyway so must not handle sanity checks.

mkarg:
With my latest commit ontop I have replaced the explicit zero-length detection by simply catching `NothingToRead`. As announced, I added a warning to the JavaDocs that this will only work with Johnzon JSON-P -- which I assume is 100% of the use cases of this MBR.

Regarding your ideas about JAX-RS I have a different view. As a long-time member of the JAX-RS team (I  had been an external advisor to JSR 311, expert group member of JSR 339 and JSR 370; now committer to Jakarta REST and publisher of JAX-RS 2.1.1 through 2.1.5) I need to tell you that this part of the JAX-RS spec actually makes *pretty much* sense. `NoContentException` can be catched by a `ReaderInterceptor`, so *the application programmer* can detect this special case and e. g. replace that exception by an entity of *his* choice, *before* the JAX-RS engine chimes in and turns the result into a `400 Bad Request`. There *are* existing applications where processing empty input streams *does* make sense, so this feature is *needed*. That "synthetic" empty entity will be propagated *to resouce methods* for regular processing of the request. This works independently of the used MBR, MIME type, and entity class. If you simply throw some MBR- or MIME-type-specific exception, the application programmer has no chance to do exactly that, as it will get propagated directly to the exception mapper, leading to 400 always -- without *any* chance for the application programmer to override this behavior. We *deliberately* introduced this interception possibility for this special case in JAX-RS 2.1. If you are unlucky with our decisions or spec, please open an issue with [our tracker on Github](https://github.com/eclipse-ee4j/jaxrs-api/issues). We are happy to explain our features, and why we invented them. :-)

rmannibucau:
Hi Markus

The warning is not needed except for jsonb provider, pkease remove it from other mbr

Also dont catch a johnzoncore exception since we can switch jsonp impl (as explained earlier)
Really think the mapper is saner and will reach your goal to support all MBR

Also i opened https://github.com/eclipse-ee4j/jaxrs-api/issues/798 since this is a spec regression (so exception mapper will enable to get back the old behavior for end users without a toggle)

mkarg:
@rmannibucau This is the point where I need to chime out. As a JAX-RS committer my position is that the spec is as it is, will most likely never get changed in the way you want it to be, so I have to take care that all MBRs either are compliant with the spec, or do not claim compliance. I understand your position, but please do all wanted changes simply ontop of my commit, as my work ends here: Compliance is given in just the described case. Please note also that the EF is strict about compliance claims, so if incompliance persists, the EF will react in some way.

rmannibucau:
@mkarg i see but not sure what to merge then since at the end the fix is just a mapper. Fo you want i merge, revert and fix? (I dont care, it is just unclear for me in current state).
Also note that we are compliant since never ran through a jaxrs tck, tomee and meecrowave would be but will be compliant with a mapper.
Finally since jaxrs broke its own spec it must fix it and it is a perfect valid case of challenge where a vendor can not respect the spec due to its inconsistencies so all good.

mkarg:
@rmannibucau We did not break our own spec. There is a difference between compliance to JAX-RS 1.0, 1.1, 2.0, 2.1, etc. We added features ontop withouth breaking the interfaces, but unless you actively support these features, you are simply compliant with e. g. 1.1 but not 2.0 or 2.0 but not 2.1. So we will definitively never "fix" that. Either you take the spec for granted, or you simply need to explicitly clarify _which version_ you fulfil. At the moment, your POM declares a JAX-RS version you do not fulfil, that's why I opened this ticket and provided a simple (but clearly slow) fix. My target is not becoming a Johnzon committer or even provide the best and fastest solution possible. I just did the sole thing I could offer you by the JAX-RS team: the code you see now. If you merge the original few commits, then you fulfil JAX-RS 2.1. If you don't then you don't even fulfil JAX-RS 2.0. If you want a performance fix, then feel free to cherry-pick some of my original commits, and put your performance solution ontop.

rmannibucau:
Perf fix is push and fine now so it is no more an issue
Yes the spec broke itself as shown in the related ticket. Spec is not only about interfaces but mainly about the behavior put behind them and this one had been broken
I will add the exception mapper tonight, is it ok to close this issue after?

mkarg:
@rmannibucau Again, we did not break the behaviour, we intentionally decided that from 1.1 to 2.0 we *want* an additional exception. If you want the old behaviour, fell free to stick with 1.1 *or* simply put a clear warning in the JavaDocs explicitly declaring intended incompliance (which is fine). In the end, it is *our* decision as JAX-RS committers to define the wanted outcome of the spec, and to decide what is a bug and what is an intended change, independent of anybody's personal opinion (which is fine, too). And *this* change was *wanted* and will stay for 2.x without any loopholes or opt-outs. You can propose this change for 3.x, but this will not help *today*: The current Johnzon code still *is* incompliant and the *sole* solution is to either not declare it compliant (i. e. explicitly note the incompliance to the MBRs) or simply fix it. It is not *my* decision how Johnzon team is handling this situation. I just told you the authoritative interpretation of the spec, and I spend you some spare time for proposals how to fix it. Now take it or leave it, it is completely in Johnzon's hands, not mine. Whether Johnzon's MBRs will be compliant after whatever announced change were actually merged is to be analyzed *after* thes changes actually happend. So I will see what team Johnzon comes up with and will tell you the authoritative decision if it *then* is compliant or not. Whether you keep the JIRA ticket / this PR open or not is not my decision, but team Johnzon's. I'm not part of that, so I don't have a vote.

rmannibucau:
@mkarg please, take a pause and check out that it broke end users. We were used to put exception mappers on parser exception, we were used to have a 500 by default, this is two trivial to reproduce use cases you broke with this change. JavaEE (jakarta now) forbids such breaking changes, period, specific expert groups didnt have the choice to change that.
Now you last sentence is wrong, you have a "vote" but what, as a project, we can't accept is to break our end users.

rmannibucau:
I just checked out the spec and seems this PR is wrong (4.2.4 from JAX-RS 2.1 Final):


    When reading zero-length message entities all pre-packaged MessageBodyReader implementations, except the JAXB one and those for the (boxed) primitive types above, MUST create a corresponding Java
    object that represents zero-length data. The pre-packaged JAXB and the pre-packaged primitive type
    MessageBodyReader implementations MUST throw a NoContentException for zero-length message
    entities.


So a few points:

1. "pre-packaged MessageBodyReader implementations" does not concern johnzon (it does TomEE, Meecrowave etc but at johnzon level it is just an user library so out of scope),
2. assuming 1 is true even for johnzon, it does not concern jaxb nor primitive so should create a zero-length data object which means throwing an exception since it is invalid for any JSON object.

So at the end our current implementation is exactly aligned on the spec even if I suspect the spec wording is not the intended one.

I will add the a way to enable to throw NoContentException if desired but will not handle it by default to respect the spec and keep backward compatibility (we really can't break already deployed API).

mkarg:
@rmannibucau Maybe I do have a vote, actually I don't know as I am not used to Apache rules. :-) We all understood your point already, no need to repeat it. JAX-RS 2.1 and JAX-RS 2.0 *are* fully backwards compatible in the sense that an application written against JAX-RS 1.1 works 100% the same on JAX-RS 2.0. Nothing breaks here, but see, there is reason why 2.0 was not simply named 1.2: Once you write a component for JAX-RS 2.x, the rules have changed, and you need to follow the new rules. And the resulting component can *only* be used by application that also follow these new rules *or stick with JAX-RS 1.x*. There is *no* guarantee in Java EE that *no changes at all* are needed, for example see CMP (Container Managed Persistency): It was simply dropped in favor of JPA, which enforced a *massive* overhaul of existing applications back then, and it broke *all* existing applications. JAX-RS 2.0 and 2.1 passed all levels of the JCP process and finally were published as official standards everbody need to follow to claim compliance. Even if you personally think this was a wrong decision by everybody involved in the standardization process, there will not be a change in the interpretation of the existing spec, only the possibility to change it in an upcoming spec (you are invited to participate in that). Also "period" ;-). BTW, it is fine if you add an option to Johnzon's MBRs that explicitly switch off JAX-RS compliance to keep backwards compatibility. We often have this situation that bugs or incompliances have to be kept, and you could follow that path, even if it means that an existing JAX-RS application has to set a Johnzon-specific config option in its bootsrap, it would be a possible solution.

mkarg:
@rmannibucau Your assumption is wrong. Please check the Javadocs of MBR class of JAX-RS 2.0 and 2.1 and you will see that my PR is *not* about that spec chapter you talk about but solely about that Javadocs. At JAX-RS there is concensus that the JavaDocs are normative, in *addition* to the spec. And no, I am not wrong in that case. And no, we do not discuss that fact. ;-)

rmannibucau:
@mkarg maybe but then spec and javadoc say opposite things so spec must be fixed before any implementation can be considered not compliant.

mkarg:
@rmannibucau Spec and JavaDoc say *additive* things, not *different* things. We wanted to fix spec and TCK before JAX-RS 2.1.6 but the Jakarta Working Group decided that no changes at all must happen and we have to do that *after* Jakarta EE 8. The changes are coming definitively, including a TCK that will claim the Johnzon MBRs as failing. BTW, the Jakarta Spec Committee decided that the final word in the interpretation of a spec and in deciding about compliance has the particular spec project (here: JAX-RS). So independent of any arguments and opinions, there is one official decision, and that is, that *for now* Johnzon's MBRs are de fact not compliant and there is *no regression* in JAX-RS as per final judgement of the JAX-RS committers.

rmannibucau:
@mkarg as per spec Johnzon MBR are standard since they are not pre-packaged....so in any case we are good. But once again it is not my worrying point, it is more that the spec broke a rule users used. You can't count the invalid incoming payload and return it in response headers now without handling NoContentException, this is per itself a regression since this code was working in JAXRS 1, and it is not the only example you can find.

rmannibucau:
it is now fixed on master, feel free to validate it before the release (guess I'll do it tomorrow).

mkarg:
@rmannibucau I will not continue the discussion about whether JAX-RS 2.x broke anything. This is years ago and the JAX-RS team as the final instance decided to treat the spec "as-is". Feel free to continue in the issue you raised regarding a change in a *future* spec. For now, all is said from both sided, no need to spend more time into that.

rmannibucau:
Guess this one is done so we can close it?

jeanouii:
Closing; Feel free to re-open if needed

PR #53, opened 2019-09-26
JOHNZON-282: add support for @JsonbTypeAdapter at class level

ArneLimburg:


johnzon-jsonb/src/test/java/org/apache/johnzon/jsonb/AdapterTest.java
```diff
@@ -16,8 +17,9 @@
  * specific language governing permissions and limitations
  * under the License.
  */
-package org.apache.johnzon.jsonb;
 
+
+import static org.hamcrest.CoreMatchers.*;
```
rmannibucau: No wildcard please ;)

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
```diff
@@ -1033,6 +1037,20 @@ private boolean isNumberType(final Type type) {
...
+        return Class.class.isInstance(type) || ParameterizedType.class.isInstance(type);
+    }
+
+    private Class<?> getRawType(final Type type) {
+        if (!Class.class.isInstance(type) && !ParameterizedType.class.isInstance(type)) {
+            throw new IllegalStateException("Unsupported generic type " + type.getClass().getName());
+        } else if (Class.class.isInstance(type)) {
+            return Class.class.cast(type);
+        } else /*if (ParameterizedType.class.isInstance(type))*/ {
+            return getRawType(ParameterizedType.class.cast(type).getRawType());
```
rmannibucau: Cant it loop?

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
```diff
@@ -868,8 +868,10 @@ public void close() throws IOException {
         ReaderConverters(final DecoratedType annotationHolder) {
             final boolean numberType = isNumberType(annotationHolder.getType());
             final boolean dateType = isDateType(annotationHolder.getType());
+            final boolean hasRawType = hasRawType(annotationHolder.getType());
             final JsonbTypeDeserializer deserializer = annotationHolder.getAnnotation(JsonbTypeDeserializer.class);
             final JsonbTypeAdapter adapter = annotationHolder.getAnnotation(JsonbTypeAdapter.class);
+            final JsonbTypeAdapter typeAdapter = hasRawType ? getRawType(annotationHolder.getType()).getDeclaredAnnotation(JsonbTypeAdapter.class) : null;
```
rmannibucau: Shouldnt it be done in the decorated type? There shouldnt be much type reflection there. Wdyt?

johnzon-jsonb/src/test/java/org/apache/johnzon/jsonb/AdapterTest.java
```diff
@@ -44,13 +47,21 @@ public void adapt() throws Exception {
...
+            assertThat(toString, containsString("\"bar\":\"1\""));
+            assertThat(toString, containsString("\"dummy\":\"2\""));
+            assertThat(toString, containsString("\"baz\":\"3\""));
+            assertThat(toString, endsWith("}"));
+            assertEquals("{\"bar\":\"1\",\"dummy\":\"2\",\"baz\":\"3\"}".length(), toString.length());
 
             final Foo read = jsonb.fromJson(toString, Foo.class);
             assertEquals(foo.bar.value, read.bar.value);
             assertEquals(foo.dummy.value, read.dummy.value);
+            assertEquals(foo.baz.value, read.baz.value);
```
rmannibucau: I would also add a test where baz is the root object too, we had issues with that on other metadata by the past.

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
ArneLimburg: Thinking about it, it neither needs recursion nor a loop since getRawType() always returns a class object

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
ArneLimburg: You mean the hasRawType and getRawType stuff? I take a look into that

johnzon-jsonb/src/test/java/org/apache/johnzon/jsonb/AdapterTest.java
ArneLimburg: I'll do

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
rmannibucau: Don't think so, Collection<Collection<Collection<....>>> ;)

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
rmannibucau: Yep, today it is not super clean for the class (it is for fields and methods) but we support meta annotations and things like that, you can have a look to findFactory in the access mode and org.apache.johnzon.jsonb.JsonbAccessMode#getAnnotation. Guess you "just" want to reuse org.apache.johnzon.jsonb.JsonbAccessMode.ClassDecoratedType somehow here.

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
ArneLimburg: This would return Collection.class

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
```diff
@@ -1033,6 +1037,20 @@ private boolean isNumberType(final Type type) {
         return Number.class.isAssignableFrom(clazz) || clazz.isPrimitive();
     }
 
+    private boolean hasRawType(final Type type) {
+        return Class.class.isInstance(type) || ParameterizedType.class.isInstance(type);
```
rmannibucau: Hmm, i get the point where java.lang.reflect will highly likely return a Class (oracle jdk guarantees it) but this comes from a reader - which is configurable in the jsonb builder - so potentially a custom PT impl, not java.reflect backed.
Any issue harnessing the check (Class.isInstance(pt.getRawType()) to avoid a ClassCastException which would be a regression for end users?

ArneLimburg: No problem with that :-)

ArneLimburg: Will not find the time to do the changes this weekend. So go ahead, if you like. Otherwise I'll do it next week.

PR #55, opened 2019-12-05
JOHNZON-293 potential high memory consumption in dateconverter

elexx:
PR for https://issues.apache.org/jira/browse/JOHNZON-293

Adds a unit test for the old `DateConverter` and rewrites the converter to use `java.time.format.DateTimeFormatter` instead of `SimpleDateFormat`.

### DIFF
--- a/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/converter/DateConverter.java
+++ b/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/converter/DateConverter.java
@@ -20,34 +20,40 @@ package org.apache.johnzon.mapper.converter;
 
 import org.apache.johnzon.mapper.Converter;
 
-import java.text.DateFormat;
-import java.text.ParseException;
-import java.text.SimpleDateFormat;
+import java.time.Instant;
+import java.time.LocalDateTime;
+import java.time.ZoneId;
+import java.time.format.DateTimeFormatter;
+import java.time.format.DateTimeParseException;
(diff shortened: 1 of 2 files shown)


rmannibucau:
Hi @elexx , DateTimeFormatter and SimpleDateFormat patterns are not 1-1 so I guess this is a breaking change. It can also be good to configure the timezone if we are breaking the API - to ensure UTC by default.
One option can be to import this https://github.com/apache/johnzon/blob/master/johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JohnzonBuilder.java#L693 natively in the mapper - and default converters - and drop it from jsonb module (except the built-in configuration/properties of jsonb but it is a few adapters).
Finally for compatibility I guess we can try {datetimeformatter}catch(e) {new SimpleDateFormat()} to ensure we don't break version and dont need a version change for such internal.

Hope it makes, sense, wdyt?

elexx:
With all the new insights (not actually a memleak, just potential "high" memusage, not backwards compatible or introducing try/catch fallbacks), I'm not sure the change is worth the effort and risk.

Maybe we just should add some Javadoc to the class, warning that it might consume a bit more memory than one might expect and provide some java.time.* Converters instead?

rmannibucau:
I am happy with whatever works for you :)

rmannibucau:
Do we want to keep it open?

elexx:
+1 for closing it. I'll also close the related Jira issue.

PR #62, opened 2020-03-20
Added possibility to use minus as array index for the remove operation.

Neoklosch:
## Problem

Currently it is not allowed to remove an entry by using a `-` as array index.
Since it is allowed for the add operation, it should also be allowed for all the other operations.
Further this is necessary if johnzon is used together with the JSON Patch library.

## Solution

I've added a check for the `-` index in the `getArrayIndex` method to return the index of the last entry.
Additionally I've made the check in the `validateArrayIndex` more explicit.

## Test(s) added 

One test that perform a remove with a `-` as array index.

### DIFF
--- a/johnzon-core/src/main/java/org/apache/johnzon/core/JsonPointerImpl.java
+++ b/johnzon-core/src/main/java/org/apache/johnzon/core/JsonPointerImpl.java
@@ -339,8 +339,7 @@ public class JsonPointerImpl implements JsonPointer {
 
             try {
                 JsonArray jsonArray = (JsonArray) jsonValue;
-                int arrayIndex = Integer.parseInt(referenceToken);
-                validateArraySize(jsonArray, arrayIndex, jsonArray.size());
+                int arrayIndex = getArrayIndex(referenceToken, jsonArray, false);
                 return jsonArray.get(arrayIndex);
             } catch (NumberFormatException e) {
                 throw new JsonException("'" + referenceToken + "' is no valid array index", e);
@@ -460,6 +459,8 @@ public class JsonPointerImpl implements JsonPointer {
     private int getArrayIndex(String referenceToken, JsonArray jsonArray, boolean addOperation) {
         if (addOperation && referenceToken.equals("-")) {
             return jsonArray.size();
+        } else if (!addOperation && referenceToken.equals("-")) {
+            return jsonArray.size() - 1;
         }
 
         validateArrayIndex(referenceToken);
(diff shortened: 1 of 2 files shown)


johnzon-core/src/main/java/org/apache/johnzon/core/JsonPointerImpl.java
```diff
@@ -339,8 +339,7 @@ private JsonValue getValue(JsonValue jsonValue, String referenceToken, int curre
 
             try {
                 JsonArray jsonArray = (JsonArray) jsonValue;
-                int arrayIndex = Integer.parseInt(referenceToken);
-                validateArraySize(jsonArray, arrayIndex, jsonArray.size());
```
rmannibucau: seems we loose that validation for number cases, no?

johnzon-core/src/test/java/org/apache/johnzon/core/JsonPointerTest.java
```diff
@@ -456,6 +456,19 @@ public void testRemoveArrayElement() {
         assertEquals("[[\"bar\",\"baz\"]]", result.toString()); // [["bar","baz"]]
     }
 
+    @Test
+    public void testRemoveLastArrayElement() {
```
rmannibucau: maybe add a test for "" too since this case is handled differently

johnzon-core/src/main/java/org/apache/johnzon/core/JsonPointerImpl.java
Neoklosch: The array size validation will be executed in the validateArraySize (https://github.com/apache/johnzon/blob/d1c6a2f62ef0bb5a798161799eae4602e20a233e/johnzon-core/src/main/java/org/apache/johnzon/core/JsonPointerImpl.java#L469-L472).

johnzon-core/src/test/java/org/apache/johnzon/core/JsonPointerTest.java
Neoklosch: If I get your question correct, there is already a test for it: https://github.com/apache/johnzon/blob/d1c6a2f62ef0bb5a798161799eae4602e20a233e/johnzon-core/src/test/java/org/apache/johnzon/core/JsonPointerTest.java#L390-L396

rmannibucau: I was more thinking about the + case (know it is not what you changed but your change moves from (a || b) to (a || (b && c)), I just want to prevent the fact it becomes ((a || b) && c) 
 - or something else - during another enhancement by a test executed in the build.
Hope it makes sense.

rmannibucau:
merged since I will start the release and don't want to hold this enhancement for a test, feel free to  add the test we spoke about in another PR

in any case, thanks a lot, this is really appreciated!

PR #63, opened 2020-05-07
JOHNZON-312

ART515:
Fix for the JOHNZON-312 issue.

### DIFF
--- a/johnzon-core/src/main/java/org/apache/johnzon/core/JsonPointerImpl.java
+++ b/johnzon-core/src/main/java/org/apache/johnzon/core/JsonPointerImpl.java
@@ -31,6 +31,7 @@ import javax.json.spi.JsonProvider;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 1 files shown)


rmannibucau:
Hi @ART515 , jira is in maintenance so will leave the comment there: is the expected result you put where TUESDAY get start=null wrong for the patch? I think it should be MONDAY.start=null right?

In terms of patch I'm experiencing an alternative impl with a boolean/state propagated to say if the nested structure must be checked for patch or not, it is a bit faster than checking all strings since it just requires to check last string then the not matching subobjects should be a boolean check only.
Let me a few minutes to see if it works.

rmannibucau:
@ART515 pushed my proposed fix on master if you can review it it would be awesome. Happy to get feedbacks compared to your fix (we can revert/rework it, nothing is in the stone yet)

rmannibucau:
Is it stil needed?

rmannibucau:
up?

jeanouii:
Fixed so closing

PR #64, opened 2020-07-03
enable Dependabot v2

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,24 @@
+#
+#  Licensed to the Apache Software Foundation (ASF) under one or more
+#  contributor license agreements.  See the NOTICE file distributed with
+#  this work for additional information regarding copyright ownership.
+#  The ASF licenses this file to You under the Apache License, Version 2.0
+#  (the "License"); you may not use this file except in compliance with
+#  the License.  You may obtain a copy of the License at
+#
+#       http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing, software
+#  distributed under the License is distributed on an "AS IS" BASIS,
+#  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#  See the License for the specific language governing permissions and
+#  limitations under the License.
+#
+version: 2
+updates:
+- package-ecosystem: maven
+  directory: "/"
+  schedule:
+    interval: daily
+    time: '04:00'
+  open-pull-requests-limit: 10


rmannibucau:
Hi,

I'm personally not that moivated by dependabot for these reasons:

1. johnzon is a fully asf project where we control and do all dependencies so each time we do a new one we upgrade it as well so dependabot will just be late in general
2. dependabot has too much false positive
3. dependabot does not handle libraries properly and tries to enforce upgrades when it shouldnt (upgrade a spec version whereas the lib must stay compatible with an older one is a common one)
4. from other ASF project which activated it I can say it makes a lot of noise which is not compensated by a real gain - in particular for the kind of project johnzon is, it is different for camel for example

What do others think?

cesarhernandezgt:
If there is a way to add a withe or a blacklist in dependabot, I think this can be useful for a couple of withe liste dependencies. Ultimately I agree that the tricky part is to find the balance between false positive and actual need it upgrades.

rmannibucau:
@cesarhernandezgt https://github.com/dependabot/feedback/issues/70#issuecomment-449377695 I guess, but not sure where you are going to with that in johnzon case since we have literally no transitive dependencies outside johnzon itself. Anything particular in mind?

cesarhernandezgt:
For example, wite list things like Junit, maven plugins, tec.
https://github.com/apache/johnzon/blob/master/pom.xml#L91

Blacklist (ignore) dependencies that like geronimo-json_1.1_spec that fall into the example you provided "(upgrade a spec version whereas the lib must stay compatible with an older one is a common one)"
https://github.com/apache/johnzon/blob/master/pom.xml#L84

rmannibucau:
@cesarhernandezgt sure we can whitelist 2-3 dependencies/plugins and we can surely drop some dependencies but overall it will be only a few. Personally I focus on transitive dependencies for the end users, rest is part of the project IMHO and updated at need for spec projects. I'm more than fine to enable dependabot but before doing it I'd like a process to ensure 1. mail don't go to dev/commit lists + 2. there is somebody to handle all these changes. Today we handle it before doing a release generally if we judge it makes sense (for ex, upgrading junit 4.12 to 4.13 does not since we wouldnt use the new features yet but if writing a test we need it then it would make sense to be clear). So to summarize *my* point, I am fine having such a automotion if there is a process associated to hit and it is not just mail sent to a sink list ;).

rmannibucau:
Guess we should close it?

rmannibucau:
seems like it does not attract anyone so will close since no action is really plan on this one until it changes later

PR #65, opened 2020-07-16
Switch to Jakarta JSON Processing API

amarkevich:
geronimo-json_1.1_spec dependency cause extra manifest entry:

Require-Capability: osgi.extender;filter:="(osgi.extender=osgi.service
 loader.registrar)",**osgi.contract;osgi.contract=JavaJSONP;filter:="(&(
 osgi.contract=JavaJSONP)(version=1.1.0))"**,osgi.ee;filter:="(&(osgi.ee
 =JavaSE)(version=1.8))"

### DIFF
--- a/johnzon-distribution/pom.xml
+++ b/johnzon-distribution/pom.xml
@@ -30,22 +30,22 @@
 
   <dependencies>
     <dependency>
-      <groupId>org.apache.geronimo.specs</groupId>
-      <artifactId>geronimo-json_1.1_spec</artifactId>
-      <version>${geronimo-jsonp.version}</version>
+      <groupId>jakarta.json</groupId>
+      <artifactId>jakarta.json-api</artifactId>
+      <version>${json-api.version}</version>
       <scope>compile</scope>
     </dependency>
     <dependency>
-      <groupId>org.apache.geronimo.specs</groupId>
-      <artifactId>geronimo-json_1.1_spec</artifactId>
-      <version>${geronimo-jsonp.version}</version>
+      <groupId>jakarta.json</groupId>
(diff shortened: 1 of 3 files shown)


rmannibucau:
Hi @amarkevich , we deliver jakarta bundles already and must keep delivering javax bundles (using classifier=jakarta for geronimo and johnzon). The contract is also desired - was a fix of a recent version to enable bnd to properly package distros.
What is the env which does not work for you?

Rergarding the patch itself, distribution and maven plugin do not impact OSGi deployments normally so change should be a noop for that regard. So only root pom diff changes something and it is in scope provided anyway so it is still about the application to bring an OSGi compliant spec jar which provides the expected contracts I think.

amarkevich:
In OSGi env the
mvn:org.apache.geronimo.specs/geronimo-json_1.1_spec/${geronimo.json.version}
required to satisfy mentioned contract; unable to use the only
mvn:jakarta.json/jakarta.json-api/${json.api.version}

rmannibucau:
@amarkevich hmm, geronimo-json is a replacement of jakarta.json-api and you must not have both. Do you use geronimo-json_1.1_spec/1.4 ? If it is about jsonb you can use geronimo-jsonb_1.0_spec/1.3 (with geronimo-json_1.1 since jakarta releases are not OSGi complete). These ones works well in Karaf and equinox standalone (bnd executable goal).

For some background, we contacted eclipse to ask them to make microprofile and jakarta spec jars OSGi compliant (with contracts in particular) but it is unlikely to happen soon so we kept geronimo as main OSGi artifacts.

amarkevich:
Ok, got it

PR #67, opened 2020-07-26
[JOHNZON-319] Map all properties supported by MapperBuilder in JohnzonBuilder

jungm:
This resolves JOHNZON-319

### DIFF
--- a/johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JohnzonBuilder.java
+++ b/johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JohnzonBuilder.java
@@ -131,9 +131,12 @@ public class JohnzonBuilder implements JsonbBuilder {
                     return it;
                 }).orElse(false);
 
-        if (config.getProperty(JsonbConfig.FORMATTING).map(Boolean.class::cast).orElse(false)) {
-            builder.setPretty(true);
-        }
+        config.getProperty(JsonbConfig.FORMATTING).map(Boolean.class::cast).ifPresent(builder::setPretty);
+        config.getProperty(AbstractJsonFactory.BUFFER_STRATEGY).map(String::valueOf).ifPresent(builder::setBufferStrategy);
+        config.getProperty(JsonParserFactoryImpl.BUFFER_LENGTH).map(Integer.class::cast).ifPresent(builder::setBufferSize);
+        config.getProperty(JsonParserFactoryImpl.MAX_STRING_LENGTH).map(Integer.class::cast).ifPresent(builder::setMaxSize);
+        config.getProperty(JsonParserFactoryImpl.SUPPORTS_COMMENTS).map(Boolean.class::cast).ifPresent(builder::setSupportsComments);
+        config.getProperty(JsonParserFactoryImpl.ENCODING).map(String::valueOf).ifPresent(builder::setEncoding);
 
         config.getProperty(PolymorphicConfig.class.getName())
                 .map(PolymorphicConfig.class::cast)
@@ -144,7 +147,7 @@ public class JohnzonBuilder implements JsonbBuilder {
                     builder.setPolymorphicDiscriminatorMapper(pc.getDiscriminatorMapper());
                     builder.setPolymorphicTypeLoader(pc.getTypeLoader());
                 });
-        config.getProperty(JsonbConfig.ENCODING).ifPresent(encoding -> builder.setEncoding(String.valueOf(encoding)));
+
         final boolean isNillable = config.getProperty(JsonbConfig.NULL_VALUES)
                 .map(it -> String.class.isInstance(it) ? Boolean.parseBoolean(it.toString()) : Boolean.class.cast(it))
                 .map(serNulls -> {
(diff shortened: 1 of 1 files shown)


johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JohnzonBuilder.java
```diff
@@ -131,9 +131,12 @@ public Jsonb build() {
...
 
-        if (config.getProperty(JsonbConfig.FORMATTING).map(Boolean.class::cast).orElse(false)) {
-            builder.setPretty(true);
-        }
+        config.getProperty(JsonbConfig.FORMATTING).map(Boolean.class::cast).ifPresent(builder::setPretty);
+        config.getProperty(AbstractJsonFactory.BUFFER_STRATEGY).map(String::valueOf).ifPresent(builder::setBufferStrategy);
+        config.getProperty(JsonParserFactoryImpl.BUFFER_LENGTH).map(Integer.class::cast).ifPresent(builder::setBufferSize);
+        config.getProperty(JsonParserFactoryImpl.MAX_STRING_LENGTH).map(Integer.class::cast).ifPresent(builder::setMaxSize);
+        config.getProperty(JsonParserFactoryImpl.SUPPORTS_COMMENTS).map(Boolean.class::cast).ifPresent(builder::setSupportsComments);
+        config.getProperty(JsonParserFactoryImpl.ENCODING).map(String::valueOf).ifPresent(builder::setEncoding);
```
rmannibucau, line 139: Must use jsonb property, not jsonp one

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JohnzonBuilder.java
```diff
@@ -131,9 +131,12 @@ public Jsonb build() {
                     return it;
                 }).orElse(false);
 
-        if (config.getProperty(JsonbConfig.FORMATTING).map(Boolean.class::cast).orElse(false)) {
-            builder.setPretty(true);
-        }
+        config.getProperty(JsonbConfig.FORMATTING).map(Boolean.class::cast).ifPresent(builder::setPretty);
+        config.getProperty(AbstractJsonFactory.BUFFER_STRATEGY).map(String::valueOf).ifPresent(builder::setBufferStrategy);
```
rmannibucau, line 135: Must use johnzon.name pattern IMHO for consistency and not leak jsonp impl properties since we support other jsonp impl and already used shortnames for other feature.

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JohnzonBuilder.java
markuung: Completely oversaw that I used the jsonp property here for the encoding, thanks!
You however marked 4 lines with this comment, all the other properties look very specific to the json-p impl. Am I missing something here?

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JohnzonBuilder.java
markuung: JohnzonBuilder#generatorConfig() and JohnzonBuilder#readerConfig() also use org.apache.johnzon.* like properties, also exposing json-p impl specific properties, so I reused those. Should those maybe be changed too?

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JohnzonBuilder.java
rmannibucau: Johnzon properties are all renamed with shorter aliases in jsonb binding ("johnzon.x"). Jsonp properties are often way too long so think it is worth aligning it on this convention.

jungm: Sure thing, but I wouldn't include this change in this PR, this is only about JOHNZON-319 afterall.
Maybe settle with the full jsonp properties for now, and change the properties to jsonb style ("johnzon.*") later on? What do you think?

rmannibucau: Hmm, first thought was why merging something we dont want and which does not match the ticket (if you map all mapper props you map them johnzon.x style) but I can miss sthg (reviewed from the phone this time).

jungm: I just don't think this PR should also contain the aliasing of the jsonp properties to jsonb-style properties. But if you think otherwise I'll include it.

rmannibucau: Hmm, any PR must not use as jsonb properties some jsonp properties by design or it should use a filter like + mapping algorithm (prefix or so) to create a kind of passthrough mode for properties. Aliases are current way to code it, subnamespace would work too. Jsonb module must also not include johnzon-core imports. in these regards this PR is not mergeable since it creates a naming inconsistency and breaks module design, this is why I mentionned it.

Thinking out loud i wouldnt map all properties for jsonp but a single "properties" (map) one and wire it in all jsonb config consumers (jaxrs providers for ex).

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JohnzonBuilder.java
rmannibucau: Yep, it shouldn't have been.
Was ok-ish in standalone/ee because it was not in the main case where jsonp impl was loaded automatically so not that visible in terms of classloading and runtime but this breaks the module design (and OSGi metadata).
It is fine to copy these strings in the builder (inline if triggered cause jsonp != null for backward compat) and we should just use a Map<String, ?> for the main case (single property, for instance "johnzon.jsonp.readerfactory.properties" and "johnzon.jsonp.generator.properties" - <jsonbimpl>.<config area>.<component>.<property> pattern?).
Will enable to handle any jsonp impl configuration and future properties addition without anything more.

jeanouii:
Up?
Still valid?

PR #68, opened 2020-11-20
JOHNZON-325 Using - on anything but add operation must throw JsonException

jeanouii:

### DIFF
--- a/johnzon-core/src/main/java/org/apache/johnzon/core/JsonPatchImpl.java
+++ b/johnzon-core/src/main/java/org/apache/johnzon/core/JsonPatchImpl.java
@@ -162,11 +162,11 @@ class JsonPatchImpl implements JsonPatch {
                    final JsonValue value) {
             this.provider = provider;
             this.operation = operation;
-            this.path = new JsonPointerImpl(provider, path);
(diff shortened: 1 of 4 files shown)


rmannibucau (changes requested):
Overall it looks good with two small details/notes:

1. String.format should likely stay concatenations (to avoid the pattern usage which is overkill there)
2. testRemoveLastArrayElement should still work OOTB so I propose to have a toggle to become spec compliant but with less features than we have some a few versions if we can't have both at the same time (happy to help on this last point if needed)

rmannibucau:
cc @Neoklosch is it ok for you to break the remove last element with "-" in pointer. I propose to move the logic to remove last element in JsonPatchImpl and not in JsonPointerImpl. This will require to extract part of JsonPointerImpl in a shared class to let the patch control the behavior of this "-" case but it will enable us to not break all our features and become spec compliant. Wdyt?

rmannibucau (approved):

Neoklosch:
Sure, being spec compliant should always be the goal of a library that implements a standard.
Therefore go for it please.
The implementation looks also fine to me.
Good job 👍

PR #69, opened 2020-12-01
JOHNZON-326 Fix for handling primitive arrays in JsonB

jgallimore:
This PR ensures that toArray() is called in JohnzonJsonB so primitive arrays are handled correctly. Tests have been added for arrays of each primitive type.


jeanouii (approved):
Looks good to me.

Good catch Jon

PR #70, opened 2020-12-02
JOHNZON-325 see if we can get something that makes everyone happy

jeanouii:
Not sure if this is what the discussion was suggesting

### DIFF
--- a/johnzon-core/src/main/java/org/apache/johnzon/core/JsonProviderImpl.java
+++ b/johnzon-core/src/main/java/org/apache/johnzon/core/JsonProviderImpl.java
@@ -57,7 +57,7 @@ public class JsonProviderImpl extends JsonProvider implements Serializable {
     /**
      * This makes Johnzon to be compliant with spec, essentially regarding the JSON Pointer /-
      */
-    private final boolean isStrictCompliance = Boolean.getBoolean("org.apache.johnzon.STRICT_JSONP_COMPLIANCE");
+    private final boolean isStrictCompliance = Boolean.getBoolean("org.apache.johnzon.strict-jsonp-compliance");
 
     private final Supplier<BufferStrategy.BufferProvider<char[]>> bufferProvider = new Cached<>(() ->
         BufferStrategyFactory.valueOf(System.getProperty(AbstractJsonFactory.BUFFER_STRATEGY, "QUEUE"))
--- a/johnzon-core/src/main/java/org/apache/johnzon/core/JsonProviderImpl.java
+++ b/johnzon-core/src/main/java/org/apache/johnzon/core/JsonProviderImpl.java
@@ -56,6 +56,10 @@ public class JsonProviderImpl extends JsonProvider implements Serializable {
 
     /**
      * This makes Johnzon to be compliant with spec, essentially regarding the JSON Pointer /-
+     *
+     * The setting is necessary for Johnzon to be compliant with JSR-374 / Jakarta JSON Processing (https://github.com/eclipse-ee4j/jsonp/),
+     * which specifies the "/-" reference points to the index after the last element in the
+     * array: https://github.com/eclipse-ee4j/jakartaee-tck/tree/master/src/com/sun/ts/tests/jsonp/api/(pointertests|patchtests)
      */
     private final boolean isStrictCompliance = Boolean.getBoolean("org.apache.johnzon.strict-jsonp-compliance");
 
--- a/johnzon-core/src/main/java/org/apache/johnzon/core/JsonProviderImpl.java
+++ b/johnzon-core/src/main/java/org/apache/johnzon/core/JsonProviderImpl.java
@@ -18,6 +18,8 @@
  */
 package org.apache.johnzon.core;
 
+import org.apache.johnzon.core.spi.JsonPointerFactory;
+
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.Reader;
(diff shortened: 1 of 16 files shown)


johnzon-core/src/main/java/org/apache/johnzon/core/JsonProviderImpl.java
```diff
@@ -53,6 +53,12 @@
 import javax.json.stream.JsonParserFactory;
 
 public class JsonProviderImpl extends JsonProvider implements Serializable {
+
+    /**
+     * This makes Johnzon to be compliant with spec, essentially regarding the JSON Pointer /-
+     */
+    private final boolean isStrictCompliance = Boolean.getBoolean("org.apache.johnzon.STRICT_JSONP_COMPLIANCE");
```
rmannibucau, line 60: a few notes on this one (hopefully from he most trivial to the design ones):

0. naming convention is aligned on tomcat but not johnzon (strict-jsonp-complance)
1. in jsonpointer constructor it should likely be converted to an int directly to avoid the if (minor)
2. makes me wonder about our default, we discussed on the list of a potentially breaking change in 1.3 but this flag as such will likely never be set and stay forever which does not sound very good to me
3. I know we already use a lot system properties but it is mainly for tuning/perf, here it is about the application behavior which means it can be set programmatically in multi app containers (I'm more thinking about OSGi than EE here) which can make it random so I'm quite mitigated
4. I dont know if you had a look but in openwebbeans we use cdi-se standalone SPI and we added a spi behind it to enable to plug another impl (owb, meecrowave, tomee, ...). We can probably do the same, ie add a JohnzonJsonPointerFactory SPI we load with the provider (plain serviceloader, and we take the one with the higheest ordinal/priority to avoid to fail if there are 2 in hierarchical classloaders). Then we can add a johnzon-jsonp-strictpointer module (if you find a better name ;)). This solves it without the need of the flag. I'm happy to help on this one if needed but think it can enable this flag without the pitfall of a global toggle like this one (in OSGi it would be linked or not, in EE it is present or not etc).

wdyt?

johnzon-core/src/main/java/org/apache/johnzon/core/JsonProviderImpl.java
```diff
@@ -53,6 +53,12 @@
 import javax.json.stream.JsonParserFactory;
 
 public class JsonProviderImpl extends JsonProvider implements Serializable {
+
+    /**
+     * This makes Johnzon to be compliant with spec, essentially regarding the JSON Pointer /-
```
jgallimore, line 58: Given that the TCK is open, I'd suggest referencing the specific test case here. Not so much for a javadoc point of view, but more to track the history. The JSR (as far as I can see) doesn't have any spec text at all, so a developer going looking for it, will at best arrive at the RFC.

The setting is necessary for Johnzon to be compliant with JSR-374 / Jakarta JSON Processing (https://github.com/eclipse-ee4j/jsonp/), which specifies the "/-" reference points to the index _after_ the last element in the array: https://github.com/eclipse-ee4j/jakartaee-tck/tree/master/src/com/sun/ts/tests/jsonp...<whichever test>

johnzon-core/src/main/java/org/apache/johnzon/core/JsonProviderImpl.java
jgallimore: 4. I did wonder the same, but then we potentially have a scenario where the app server needs the setting to pass that TCK, and the setting isn't available in when using Johnzon standalone - you then can't get the same behaviour in both. Unit-testing is a use case where I could see this being useful, for example.

While I'm sure we could solve this in the app server, I think my own preference would be to enable the behaviour in Johnzon itself.

jeanouii: > a few notes on this one (hopefully from he most trivial to the design ones):
> 
> 1. naming convention is aligned on tomcat but not johnzon (strict-jsonp-complance)

Done

> 2. in jsonpointer constructor it should likely be converted to an int directly to avoid the if (minor)

Well in JsonPointer it's an int, but it might be something else in the future and something else for other objects. So I'm passing the flag and let the objects to handle it.

> 3. makes me wonder about our default, we discussed on the list of a potentially breaking change in 1.3 but this flag as such will likely never be set and stay forever which does not sound very good to me

We are in 1.2.9-SNAPSHOT so default must be backward compatible as far as I understood from the discussion.
But in essence, you are right, aside from having the stamp, no one will really care about it.

> 4. I know we already use a lot system properties but it is mainly for tuning/perf, here it is about the application behavior which means it can be set programmatically in multi app containers (I'm more thinking about OSGi than EE here) which can make it random so I'm quite mitigated
> 5. I dont know if you had a look but in openwebbeans we use cdi-se standalone SPI and we added a spi behind it to enable to plug another impl (owb, meecrowave, tomee, ...). We can probably do the same, ie add a JohnzonJsonPointerFactory SPI we load with the provider (plain serviceloader, and we take the one with the higheest ordinal/priority to avoid to fail if there are 2 in hierarchical classloaders). Then we can add a johnzon-jsonp-strictpointer module (if you find a better name ;)). This solves it without the need of the flag. I'm happy to help on this one if needed but think it can enable this flag without the pitfall of a global toggle like this one (in OSGi it would be linked or not, in EE it is present or not etc).

Don't mind but not sure I can address this now though.
And as mentioned in 3/ no one will use this property unless you want to be certified. So can be seen as tuning/perf.

> 
> wdyt?

rmannibucau:
Hopefully making it clearer here is the toggle proposal I mentionned: https://github.com/apache/johnzon/pull/71

jeanouii:
I merged your PR into mine so everything should now be consistent and fixed

rmannibucau (approved):

PR #72, opened 2020-12-02
[JOHNZON-327] Json Schema Validation: added integer type validation.

manuelrazola:
Hi!
As per the official json-schema specification, integer type validation is missing from the current implementation.
http://json-schema.org/understanding-json-schema/reference/numeric.html

This PR tries to tackle that in the least intrusive way possible.
Thanks in advance!

### DIFF
--- a/johnzon-jsonschema/src/main/java/org/apache/johnzon/jsonschema/spi/builtin/IntegerValidation.java
+++ b/johnzon-jsonschema/src/main/java/org/apache/johnzon/jsonschema/spi/builtin/IntegerValidation.java
@@ -7,6 +7,7 @@ import java.util.stream.Stream;
 import javax.json.JsonNumber;
 import javax.json.JsonString;
 import javax.json.JsonValue;
+import javax.json.JsonValue.ValueType;
 
 import org.apache.johnzon.jsonschema.ValidationResult;
 import org.apache.johnzon.jsonschema.ValidationResult.ValidationError;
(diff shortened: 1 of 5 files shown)


johnzon-jsonschema/src/main/java/org/apache/johnzon/jsonschema/spi/builtin/IntegerValidation.java
```diff
@@ -0,0 +1,50 @@
...
+		private Impl(final String pointer, final Function<JsonValue, JsonValue> valueProvider) {
+			super(pointer, valueProvider, JsonValue.ValueType.NUMBER);
+		}
+
+		@Override
+		protected Stream<ValidationError> onNumber(JsonNumber number) {
+			final double value = number.doubleValue();
+			if (value % 1 == 0) {
+				return Stream.empty();
+			} else {
```
rmannibucau, line 38: you can drop the else (as keyword and directly do "return Stream.of(new ValidationResult.ValidationError(pointer, "Expected integer but got " + value));") since previous block uses "return"

johnzon-jsonschema/src/main/java/org/apache/johnzon/jsonschema/spi/builtin/IntegerValidation.java
```diff
@@ -0,0 +1,50 @@
...
+
+	private static class Impl extends BaseValidation {
+
+		private Impl(final String pointer, final Function<JsonValue, JsonValue> valueProvider) {
+			super(pointer, valueProvider, JsonValue.ValueType.NUMBER);
+		}
+
+		@Override
+		protected Stream<ValidationError> onNumber(JsonNumber number) {
+			final double value = number.doubleValue();
```
rmannibucau, line 35: hmm, wonder if it shouldn't be merged with multipleof impl: org.apache.johnzon.jsonschema.spi.builtin.MultipleOfValidation
we can image make org.apache.johnzon.jsonschema.spi.builtin.MultipleOfValidation.Impl visible in the package and create method return a new MultipleOfValidation.Impl(...., 1) instead of pseudo-duplicating it, would likely be more explicit we reuse this logic as impl.
what do you think?

johnzon-jsonschema/src/test/java/org/apache/johnzon/jsonschema/JsonSchemaValidatorTest.java
```diff
@@ -355,6 +355,25 @@ public void exclusiveMaximum() {
...
+                        .add("age", jsonFactory.createObjectBuilder()
+                                .add("type", "integer")
+                                .build())
+                        .build())
+                .build());
+
+        assertTrue(validator.apply(jsonFactory.createObjectBuilder().build()).isSuccess());
+        assertTrue(validator.apply(jsonFactory.createObjectBuilder().add("age", 30).build()).isSuccess());
+        // check no decimal numbers allowed
+        assertFalse(validator.apply(jsonFactory.createObjectBuilder() .add("age", 30.3) .build()).isSuccess());
```
rmannibucau, line 373: we need to test with BigInteger and BigDecimal too I guess

rmannibucau (changes requested):
Overall it is a very good addition IMO, a few details proposals to simplify it/align it on other impls but looks pretty good to me.

johnzon-jsonschema/src/main/java/org/apache/johnzon/jsonschema/spi/builtin/IntegerValidation.java
manuelrazola: Possibly, will test how it would look like and come back tomorrow ;)

manuelrazola: Implemented as requested

johnzon-jsonschema/src/test/java/org/apache/johnzon/jsonschema/JsonSchemaValidatorTest.java
manuelrazola: Updated tests

PR #74, opened 2021-07-09
True object and array streaming

csaboka:
As the comments say, `getArrayStream()` and `getObjectStream()` were very simplistic and didn't conform to the spirit of the specification, which requires lazy loading from the underlying stream.

Now these methods should behave lazily, only keeping the minimal amount of data in memory that's necessary for the stream.

### DIFF
--- a/johnzon-core/src/main/java/org/apache/johnzon/core/JohnzonJsonParserImpl.java
+++ b/johnzon-core/src/main/java/org/apache/johnzon/core/JohnzonJsonParserImpl.java
@@ -187,6 +187,10 @@ public abstract class JohnzonJsonParserImpl implements JohnzonJsonParser {
                 return false;
             }
 
+            if (next != Event.KEY_NAME) {
+                throw new IllegalStateException("Expected key name event but got " + next + " instead.");
+            }
+
             String key = parser.getString();
             parser.next();
             JsonValue value = parser.getValue();
--- a/johnzon-core/src/main/java/org/apache/johnzon/core/JohnzonJsonParserImpl.java
+++ b/johnzon-core/src/main/java/org/apache/johnzon/core/JohnzonJsonParserImpl.java
@@ -154,10 +154,11 @@ public abstract class JohnzonJsonParserImpl implements JohnzonJsonParser {
 
                 if (next == Event.END_ARRAY) {
                     return false;
-                } else {
-                    action.accept(getValue());
-                    return true;
                 }
+
+                action.accept(getValue());
+                return true;
+
             }
         };
 
@@ -177,16 +178,17 @@ public abstract class JohnzonJsonParserImpl implements JohnzonJsonParser {
             @Override
             public boolean tryAdvance(Consumer<? super Entry<String, JsonValue>> action) {
                 Event next = next();
-                
+
                 if (next == Event.END_OBJECT) {
                     return false;
-                } else {
-                    String key = getString();
-                    next();
-                    JsonValue value = getValue();
-                    action.accept(new AbstractMap.SimpleImmutableEntry<>(key, value));
-                    return true;
                 }
+
+                String key = getString();
+                next();
+                JsonValue value = getValue();
+                action.accept(new AbstractMap.SimpleImmutableEntry<>(key, value));
+                return true;
+
             }
         };
 
--- a/johnzon-core/src/main/java/org/apache/johnzon/core/JohnzonJsonParserImpl.java
+++ b/johnzon-core/src/main/java/org/apache/johnzon/core/JohnzonJsonParserImpl.java
@@ -138,6 +138,28 @@ public abstract class JohnzonJsonParserImpl implements JohnzonJsonParser {
         }
     }
 
+    private static class ArrayStreamSpliterator extends Spliterators.AbstractSpliterator<JsonValue> {
+
+        private final JohnzonJsonParserImpl parser;
+
+        ArrayStreamSpliterator(JohnzonJsonParserImpl parser) {
+            super(Long.MAX_VALUE, Spliterator.IMMUTABLE | Spliterator.NONNULL | Spliterator.ORDERED);
+            this.parser = parser;
+        }
+
+        @Override
+        public boolean tryAdvance(Consumer<? super JsonValue> action) {
+            Event next = parser.next();
+
+            if (next == Event.END_ARRAY) {
+                return false;
+            }
+
+            action.accept(parser.getValue());
+            return true;
+        }
+    }
+
     @Override
     public Stream<JsonValue> getArrayStream() {
         Event current = current();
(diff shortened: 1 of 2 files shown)


johnzon-core/src/main/java/org/apache/johnzon/core/JohnzonJsonParserImpl.java
```diff
@@ -134,18 +140,57 @@ public void skipArray() {
...
-        //X TODO this implementation is very simplistic
-        //X I find it unintuitive what the spec intends here
-        //X we probably need to improve this
-        return getArray().stream();
+        Event current = current();
+        if (current != Event.START_ARRAY) {
+            throw new IllegalStateException(current + " doesn't support getArrayStream()");
+        }
+
+        Spliterator<JsonValue> arraySpliterator = new Spliterators.AbstractSpliterator<JsonValue>(
```
rmannibucau, line 148: What about making it an inner private class reused in both locations (private static class)?

johnzon-core/src/main/java/org/apache/johnzon/core/JohnzonJsonParserImpl.java
```diff
@@ -134,18 +140,57 @@ public void skipArray() {
...
+                Long.MAX_VALUE, Spliterator.IMMUTABLE | Spliterator.NONNULL | Spliterator.ORDERED) {
+
+            @Override
+            public boolean tryAdvance(Consumer<? super JsonValue> action) {
+                Event next = next();
+
+                if (next == Event.END_ARRAY) {
+                    return false;
+                } else {
```
rmannibucau, line 157: no need of else since first branch "returns"

rmannibucau (changes requested):
2 small picky details to update to get it perfectly but overall looks  good
would be great to add error cases too in tests to ensures it behaves properly (getObjectStream does not test key event properly for example so error should be more precise than the one we would get right now IMHO)

johnzon-core/src/main/java/org/apache/johnzon/core/JohnzonJsonParserImpl.java
csaboka: The generic arguments and the `tryAdvance()` implementations are different, so what could be reused? As far as I can see, the only thing duplicated between the two Spliterator implementations are the arguments passed to the superclass constructor. I can extract that to a helper class if you'd like, but it feels overkill to me.

csaboka (commented):
> would be great to add error cases too in tests to ensures it behaves properly (getObjectStream does not test key event properly for example so error should be more precise than the one we would get right now IMHO)

Can you clarify this a bit? Should I check against the result of the first `next()` call in `getObjectStream` to make sure it returned `KEY_NAME`? As far as I can tell there is no need to check the value of the second call because `getValue` will complain anyway if the parser is in the wrong state.

rmannibucau:
@csaboka it is mainly about ensuring the error message references it is about the key and not just a "string" (which can be a value), think we should harness that

johnzon-core/src/main/java/org/apache/johnzon/core/JohnzonJsonParserImpl.java
rmannibucau: if you think you need 2 inner classes it is fine for me too, what I want to ensure is that it is not referencing $0 (this).
it is also fine ot inject the impl of tryAdvance as a lambda (the compiler will make it right normally) so maybe something along:

> new SimpleSplitIterator(action -> {  Event next = next(); ..... });

this enables to factorize the config and avoids the $0 ref.

csaboka: It needs to reference the enclosing parser instance, though, because it is calling methods on that instance like `next()`. If `tryAdvance` is factored out into a lambda, it will still capture `this`. As far as I can tell, a lambda won't help anything in this situation. It either needs to stay a nonstatic class like it is now, or it can be turned into a static class, but instances of that static class will need to explicitly reference the parser they came from.

rmannibucau: Pass parser as a param ;)
Lambdas have 3 conversions so if well refactored it solves it.
You can use javap to ensure you reached that goal

csaboka: I've pushed some changes. Is this what you had in mind?

csaboka:
> @csaboka it is mainly about ensuring the error message references it is about the key and not just a "string" (which can be a value), think we should harness that

I've added a check to make sure we get `KEY_NAME` in the object stream (unless it's `END_OBJECT` and we're done). I couldn't come up with a good test case for it, though. All the cases I could think of where `next()` would return something other than `KEY_NAME` ended up failing with a `JsonParsingException` instead.

rmannibucau:
While consistent with parser api it is fine, just needs to be harnessed to ensure this consistency.
If you cant get a more precise message it is not critical for now but test should be there IMHO.

Side note: ill be off computer for some time so if somebody wants to merge once the test is added consider you got my +1

csaboka:
Added some unit tests to validate that a parse exception is thrown when trying to traverse the stream and invalid JSON is encountered. I still can't figure out a good way to make the `KEY_NAME` check to fail. I could write a new subclass of JohnzonJsonParserImpl just to make it return unexpected values from `next()`, but that seems to be overkill for testing this edge case.

johnzon-core/src/test/java/org/apache/johnzon/core/JsonParserStreamingTest.java
```diff
@@ -121,4 +127,59 @@ private String parserAndConcat(String json) {
...
+        expectedMappings.put("something", "else");
+        assertEquals(expectedMappings, mappings);
+    }
+
+    @Test(expected = JsonParsingException.class)
+    public void testParseErrorInGetObjectStream() {
+        StringReader sr = new StringReader("{42}");
+        JsonParser jsonParser = Json.createParser(sr);
+
+        JsonParser.Event firstEvent = jsonParser.next();
```
rmannibucau, line 179: guess we should use a try with resource pattern to close the parser (it is required even for string readers due to memory management, even if not critical in tests it is always better to not show antipatterns in tests since they are kind of part of the doc)
if you test the message it looks ok.

But looks sufficient for me as coverage.

csaboka: None of the other tests closed the parser, that's why I assumed it's fine. I've changed now so the parser is closed not just in the new tests, but the existing ones as well.

rmannibucau (approved):

PR #76, opened 2021-11-23
[JOHNZON-355]: Find class level Adapter even when used in collection.

ArneLimburg:


johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
```diff
@@ -18,62 +18,16 @@
  */
 package org.apache.johnzon.jsonb;
 
-import org.apache.johnzon.core.Types;
```
rmannibucau, line 21: guess we can avoid this diff part ;)

rmannibucau (changes requested):
Except the header change I think we should avoid it looks perfect!
Thanks a lot!

PR #77, opened 2021-12-15
fix using patternProperties in nested objects

radu-almasan:


rmannibucau:
Hi,

Looks fishy to have validable used for some checks and root for the final one, did you refine the exact case?

rmannibucau:
up?

rmannibucau:
@radu-almasan normally no, `validable` is really the value to validate extractor result, if you pass root you break all the cases not using root so part of the validations.

radu-almasan:
Regarding the 2nd half of your statement, the existing tests passed successfully and the one I added was probably specific to this case.

It's been too long since I coded this and don't remember all the details. If the PR doesn't make sense anymore we can close it.

rmannibucau:
Hi @radu-almasan ,

I took time to check this case and as suspected it is still a bug but here what I meant for the fix (`validable` is the one to handle, root is just a particular case and if it has multiple validations we should ensure the extractor works as fast as possible to avoiding the chaining from `root` is likely sane so it is mainly about rewriting provider):

```diff
index 14103bc..d62752f 100644
--- a/johnzon-jsonschema/src/main/java/org/apache/johnzon/jsonschema/JsonSchemaValidatorFactory.java
+++ b/johnzon-jsonschema/src/main/java/org/apache/johnzon/jsonschema/JsonSchemaValidatorFactory.java
@@ -197,17 +197,20 @@ public class JsonSchemaValidatorFactory implements AutoCloseable {
                             final Predicate<CharSequence> pattern = regexFactory.get().apply(obj.getKey());
                             final JsonObject currentSchema = obj.getValue().asJsonObject();
                             // no cache cause otherwise it could be in properties
-                            return (Function<JsonValue, Stream<ValidationResult.ValidationError>>) validable -> {
+                            return (Function<JsonValue, Stream<ValidationResult.ValidationError>>) root -> {
+                                final JsonValue validable = Optional.ofNullable(valueProvider)
+                                        .map(provider -> provider.apply(root))
+                                        .orElse(root);
                                 if (validable.getValueType() != JsonValue.ValueType.OBJECT) {
                                     return Stream.empty();
                                 }
                                 return validable.asJsonObject().entrySet().stream()
                                         .filter(e -> pattern.test(e.getKey()))
-                                        .flatMap(e -> {
-                                            final String[] subPath = Stream.concat(Stream.of(path), Stream.of(e.getKey())).toArray(String[]::new);
-                                            final Function<JsonValue, JsonValue> provider = new ChainedValueAccessor(valueProvider, e.getKey());
-                                            return buildValidator(subPath, currentSchema, provider).apply(validable);
-                                        });
+                                        .flatMap(e -> buildValidator(
+                                                Stream.concat(Stream.of(path), Stream.of(e.getKey())).toArray(String[]::new),
+                                                currentSchema,
+                                                o -> o.asJsonObject().get(e.getKey()))
+                                                .apply(validable));
                             };
                         })
                         .collect(toList()))
```

PR #82, opened 2022-03-16
enhance OSGi support

stbischof:
* use newer bnd-maven-plugin
* remove older felix-bundle-plugin
* remove <packaging>bundle</packaging>
* use @ServiceConsumer and @ServiceProvider annotations for better
Manifest - `Require-Capability` and `Provide-Capability`
* load BufferStrategy also over ServiceLoader
* do not export `internal` packages

@rotty3000 could you please review?

### DIFF
--- a/johnzon-core/pom.xml
+++ b/johnzon-core/pom.xml
@@ -27,7 +27,6 @@
 
   <artifactId>johnzon-core</artifactId>
   <name>Johnzon :: Core</name>
-  <packaging>bundle</packaging>
 
   <properties>
     <staging.directory>${project.parent.reporting.outputDirectory}</staging.directory>
@@ -36,14 +35,8 @@
   <build>
     <plugins>
       <plugin>
-        <groupId>org.apache.felix</groupId>
-        <artifactId>maven-bundle-plugin</artifactId>
-        <configuration>
-          <instructions>
-            <Require-Capability>osgi.extender;filter:="(osgi.extender=osgi.serviceloader.registrar)"</Require-Capability>
-            <Provide-Capability>osgi.serviceloader;osgi.serviceloader=javax.json.spi.JsonProvider</Provide-Capability>
-          </instructions>
-        </configuration>
+        <groupId>biz.aQute.bnd</groupId>
+        <artifactId>bnd-maven-plugin</artifactId>
       </plugin>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
@@ -52,6 +45,15 @@
       </plugin>
     </plugins>
   </build>
+  
+  <dependencies>
+    <dependency>
+      <groupId>biz.aQute.bnd</groupId>
+      <artifactId>biz.aQute.bnd.annotation</artifactId>
+      <version>${bnd.version}</version>
+      <scope>provided</scope>
+    </dependency>
+  </dependencies>
 
 <!-- avoid warnings at build time + this is not sufficient to run anyway (edit ts.jte)
   <profiles>
--- a/johnzon-core/src/main/java/org/apache/johnzon/core/BufferStrategyFactory.java
+++ b/johnzon-core/src/main/java/org/apache/johnzon/core/BufferStrategyFactory.java
@@ -19,14 +19,20 @@ package org.apache.johnzon.core;
 import java.util.HashMap;
 import java.util.Locale;
 import java.util.Map;
+import java.util.ServiceLoader;
 import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.stream.StreamSupport;
 
 import org.apache.johnzon.core.util.ClassUtil;
 
+import aQute.bnd.annotation.Resolution;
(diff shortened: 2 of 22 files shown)


rmannibucau:
Hi @stbischof ,

when we discussed OSGi support we decided to:

1. use maven bundle plugin and not bnd plugin
2. not put any OSGi dependency/code in all modules but OSGi (all being done through maven plugin)

About internal packages, I fear we need to export them since some integration use them - in particular websocket module (they are probably misnamed for OSGi users).

About BufferStrategy SPI usage: it is unrelated to OSGi so likely belongs to a dedicated ticket but would also need an use case I think since it is not recommended to change it (even a few default ones are for very particular cases so not sure which case we don't already cover and I'd like to avoid a SPI if not needed). Also note that the SPI usage of JSON-P - including johnzon impl - is through the maps passed to factories, not another ServiceLoader after the provider loading so if there is a need it belongs to props IMHO.

Hope it makes sense.

stbischof:
Thank you for the information and the fast reaction.

I could not understand the point where build time evaluated annotations and there dependencies are a problem.

The BufferStrategy over ServiceLoader mechanism was just because loading classes by name from a classloader in this way doesn't convince me.

But. It's your project. If there's any point where you need more OSGi support. You can ping me.

rmannibucau:
> I could not understand the point where build time evaluated annotations and there dependencies are a problem.

Idea is to not add OSGi in all modules - since most of the contributors are not OSGi experts ;) - and keep it located in a dedicated module to avoid unexpected changes.

> The BufferStrategy over ServiceLoader mechanism was just because loading classes by name from a classloader in this way doesn't convince me.

This is something we can discuss on the mailing list but long story short it is to fit the way to configure JSON-P through properties (per factory instance). ServiceLoader would be per classloader and require a priority so you wouldn't be able to use different strategies per JSON-P component instance whereas we are today. But once again, we can refine it if there is an use case but it is unrelated to OSGi ;).

> If there's any point where you need more OSGi support.

Can you refine it? We run well in OSGi - we are commonly used with aries and Karaf -  so would be great to have more details about what you miss?

stbischof:
My goal was to keep it simple for all non-osgi developers.

To only work with 2 Annotations (@ServiceProvider, @ServiceConsumer) and get everything else for free.
This Annotations only generate Entrys for the Manifest.

```
Provide-Capability                      osgi.service;objectClass:List<String>="javax.json.spi.JsonProvider";service.vendor="org.apache.johnzon";effective:=active
                                        osgi.serviceloader;osgi.serviceloader="javax.json.spi.JsonProvider";register:="org.apache.johnzon.core.JsonProviderImpl";service.vendor="org.apache.johnzon"
Require-Capability                      osgi.contract;osgi.contract=JavaJSONP;filter:="(&(osgi.contract=JavaJSONP)(version=1.1.0))"
                                        osgi.ee;filter:="(&(osgi.ee=JavaSE)(version=1.8))"
                                        osgi.extender;filter:="(&(osgi.extender=osgi.serviceloader.processor)(version>=1.0.0)(!(version>=2.0.0)))";resolution:=optional
                                        osgi.extender;filter:="(&(osgi.extender=osgi.serviceloader.registrar)(version>=1.0.0)(!(version>=2.0.0)))"
                                        osgi.serviceloader;filter:="(osgi.serviceloader=org.apache.johnzon.core.BufferStrategy)";osgi.serviceloader="org.apache.johnzon.core.BufferStrategy";resolution:=optional
                                        osgi.serviceloader;filter:="(osgi.serviceloader=org.apache.johnzon.core.spi.JsonPointerFactory)";osgi.serviceloader="org.apache.johnzon.core.spi.JsonPointerFactory";resolution:=optional

```
This isn't something, someone should care by hand. ( also not in the felix-plugin).

And at the long term. I think that the `bnd-maven-plugin` replaces `felix-plugin`.

rmannibucau:
> To only work with 2 Annotations (https://github.com/Serviceprovider, @ServiceConsumer) and get everything else for free.

This is where the issue starts, you mix two string config locations: code and pom so IMHO it is saner to stick to pom only. Other point is that most dev will just drop these useless (for them) annotations until there is an IT which fails and they will not be able to solve it, so keeping it outside code is generally saner from my past experience (but I understand your OSGi first idea which works well for pure OSGi projects).

> This isn't something, someone should care by hand. ( also not in the felix-plugin).

Have to admit it is 1-1 with the annotation as of today for me except if you want the full OSGi vision you need to review code+pom (or build) whereas currently we have it in the pom in a single config so think it is maybe just a habit point?

> And at the long term. I think that the bnd-maven-plugin replaces felix-plugin.

I don't think so, most of the Karaf ecosystem relies on felix plugin and will likely continue AFAIK since it encapsulates a lot of things, enables to use bnd advanced features but also simplifies several things on top of it.

maybe @jbonofre can confirm this point?

rmannibucau:
guess this one is done now we have an osgi module?

jbonofre:
Yes I think so.

jeanouii:
Closing because it's done

PR #89, opened 2022-04-28
Ensure Snippet works as intended for other JsonGenerator implementations

dblevins:
If our intent is to optimize for small snippets and support JSONP implementations aside from Johnzon, we still need to fall back to doing the `flush()`.

The only reason we were able to get rid of the `flush()` calls is because we added `Buffered` inside Johnzon.  Other JSON-P implementations won't understand that interface and are likely to have their own buffering.  In those situations we'll need to flush to ensure the SnippetWriter `terminate()` calls work.  It's only for Johnzon that we can skip them.

### DIFF
--- a/johnzon-core/src/main/java/org/apache/johnzon/core/Snippet.java
+++ b/johnzon-core/src/main/java/org/apache/johnzon/core/Snippet.java
@@ -30,6 +30,7 @@ import java.io.IOException;
 import java.io.Writer;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.function.Supplier;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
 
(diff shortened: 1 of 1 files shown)


rmannibucau:
@dblevins idea was to use the full string representation without johnzon (check the SnippetFactory used). It is ok to be in degraded mode for this case with other implementations. Ultimately if you want to enhance more this case you also need to change the SnippetFactory for this case.

rmannibucau:
guess this one is ok now? should we close?

jeanouii:
Does ok mean already merged in another PR so we can close without merge?
Or does it mean, it's ready to be merged

rmannibucau:
@jeanouii AFAIK we did the work in the merged PR so think we can close

PR #90, opened 2022-04-30
Catalog of deserialization error messages as they are now

dblevins:
Submitting this as its own PR with the idea that actual changes to exception messages can happen in other PRs over time.  We'll be able to see in those PRs a diff of what the message was and what it will be if the PR is merged.  Seeing the error messages we have can also help us see where things are weakest and prioritize improving that exception handling.


dblevins:
This should be closed and not merged when #91 is merged.

jeanouii:
As requested closing because #91 has been merged already

PR #91, opened 2022-05-03
Exception message improvements

dblevins:
Addresses:

 - [JOHNZON-370](https://issues.apache.org/jira/browse/JOHNZON-370) Include class, property, type and json value in mapping exceptions
 - [JOHNZON-371](https://issues.apache.org/jira/browse/JOHNZON-371) NullPointerException thrown when null value present in primitive arrays
 - [JOHNZON-372](https://issues.apache.org/jira/browse/JOHNZON-372) Entire json value serialized in exception

See PR #90 for a view of what the exceptions look like as they are currently in 1.2.18

### DIFF
--- a/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/ExceptionMessages.java
+++ b/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/ExceptionMessages.java
@@ -27,7 +27,7 @@ import static javax.json.JsonValue.ValueType.OBJECT;
 import static javax.json.JsonValue.ValueType.STRING;
 import static javax.json.JsonValue.ValueType.TRUE;
 
-public class ExceptionMessages {
+public final class ExceptionMessages {
 
     private ExceptionMessages() {
     }
--- a/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/ExceptionMessages.java
+++ b/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/ExceptionMessages.java
@@ -17,8 +17,10 @@
 package org.apache.johnzon.mapper;
 
 import javax.json.JsonValue;
+import java.lang.reflect.GenericArrayType;
 import java.lang.reflect.ParameterizedType;
 import java.lang.reflect.Type;
+import java.lang.reflect.WildcardType;
 
 import static javax.json.JsonValue.ValueType.ARRAY;
 import static javax.json.JsonValue.ValueType.FALSE;
@@ -37,23 +39,37 @@ public final class ExceptionMessages {
             final Class<?> clazz = (Class<?>) type;
             return clazz.getSimpleName();
         }
+        
         if (type instanceof ParameterizedType) {
             final ParameterizedType parameterizedType = (ParameterizedType) type;
-            final StringBuilder sb = new StringBuilder();
-            sb.append(simpleName(parameterizedType.getRawType()));
-            sb.append("<");
-
-            final Type[] args = parameterizedType.getActualTypeArguments();
-            for (int i = 0; i < args.length; i++) {
-                final Type arg = args[i];
-                sb.append(simpleName(arg));
-                if (i < args.length - 1) {
-                    sb.append(",");
-                }
-            }
-            sb.append(">");
-            return sb.toString();
+            return simpleName(parameterizedType.getRawType()) +
+                    "<" +
+                    join(",", parameterizedType.getActualTypeArguments()) +
+                    ">";
         }
+        
+        if (type instanceof GenericArrayType) {
+            final GenericArrayType genericArrayType = (GenericArrayType) type;
+            return simpleName(genericArrayType.getGenericComponentType()) + "[]";
+        }
+
+        if (type instanceof WildcardType) {
+            final WildcardType wildcardType = (WildcardType) type;
+
+            if (wildcardType.getLowerBounds().length > 0) {
+                return "? super " + join(" & ", wildcardType.getLowerBounds());
+            }
+
+            if (wildcardType.getUpperBounds().length > 0) {
+                return "? extends " + join(" & ", wildcardType.getUpperBounds());
+            }
+
+            // This should never happen, but it's always safe to fallback on getTypeName()
+            // so we do that instead of throwing an exception
+            return wildcardType.getTypeName();
+        }
+
+        // Some Type derivative we've never seen.  Fallback on getTypeName()
         return type.getTypeName();
     }
 
@@ -73,5 +89,20 @@ public final class ExceptionMessages {
         }
 
         return "json value";
+    }
+
+    private static String join(final String delimiter, final Type[] args) {
+        if (args.length == 0) {
+            return "";
+        }
+
+        final StringBuilder sb = new StringBuilder(simpleName(args[0]));
+
+        for (int i = 1; i < args.length; i++) {
+            sb.append(delimiter);
+            sb.append(simpleName(args[i]));
+        }
+
+        return sb.toString();
     }
 }
--- a/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/ExceptionMessages.java
+++ b/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/ExceptionMessages.java
@@ -1,0 +1,77 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ */
+package org.apache.johnzon.mapper;
+
+import javax.json.JsonValue;
+import java.lang.reflect.ParameterizedType;
+import java.lang.reflect.Type;
+
+import static javax.json.JsonValue.ValueType.ARRAY;
+import static javax.json.JsonValue.ValueType.FALSE;
+import static javax.json.JsonValue.ValueType.NUMBER;
+import static javax.json.JsonValue.ValueType.OBJECT;
+import static javax.json.JsonValue.ValueType.STRING;
+import static javax.json.JsonValue.ValueType.TRUE;
+
+public class ExceptionMessages {
+
+    private ExceptionMessages() {
+    }
+
+    public static String simpleName(final Type type) {
+        if (type instanceof Class) {
+            final Class<?> clazz = (Class<?>) type;
+            return clazz.getSimpleName();
+        }
+        if (type instanceof ParameterizedType) {
+            final ParameterizedType parameterizedType = (ParameterizedType) type;
+            final StringBuilder sb = new StringBuilder();
+            sb.append(simpleName(parameterizedType.getRawType()));
+            sb.append("<");
+
+            final Type[] args = parameterizedType.getActualTypeArguments();
+            for (int i = 0; i < args.length; i++) {
+                final Type arg = args[i];
+                sb.append(simpleName(arg));
+                if (i < args.length - 1) {
+                    sb.append(",");
+                }
+            }
+            sb.append(">");
+            return sb.toString();
+        }
+        return type.getTypeName();
+    }
+
+    public static String description(final JsonValue value) {
+        return description(value == null ? null : value.getValueType());
+    }
+
+    public static String description(final JsonValue.ValueType type) {
+        if (type == OBJECT || type == ARRAY || type == STRING) {
+            return "json " + type.toString().toLowerCase() + " value";
+        }
+        if (type == NUMBER) {
+            return "json numeric value";
+        }
+        if (type == TRUE || type == FALSE) {
+            return "json boolean value";
+        }
+
+        return "json value";
+    }
+}
--- a/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
+++ b/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
@@ -346,7 +346,9 @@ public class MappingParserImpl implements MappingParser {
             }
         }
         if (classMapping == null) {
-            throw new MapperException("Can't map JSON Object to " + type + ": " + config.getSnippet().of(object));
+            final String snippet = config.getSnippet().of(object);
+            final String description = ExceptionMessages.description(object);
+            throw new MapperException("Unable to map " + description + " to " + type + ": " + snippet);
         }
 
         if (applyObjectConverter && classMapping.reader != null && (skippedConverters == null || !skippedConverters.contains(type))) {
--- a/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
+++ b/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
@@ -383,12 +383,11 @@ public class MappingParserImpl implements MappingParser {
             } else {
                 t = classMapping.factory.create(createParameters(classMapping, object, jsonPointer));
             }
+        } catch (FactoryCreateException e){
+            throw e;
         } catch (Exception e) {
-            final String message = String.format("%s cannot be constructed to deserialize %s: %s%n%s",
-                    ExceptionMessages.simpleName(type), ExceptionMessages.description(object),
-                    config.getSnippet().of(object), e.getMessage()
-            );
-            throw new MapperException(message, e);
+            final String snippet = config.getSnippet().of(object);
+            throw new FactoryCreateException(type, object, snippet, e);
         }
 
         // store the new object under it's jsonPointer in case it gets referenced later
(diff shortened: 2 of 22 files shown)


johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
```diff
@@ -399,36 +403,43 @@ private Object buildObject(final Type inType, final JsonObject object, final boo
...
-                final Object convertedValue = toValue(
-                        existingInstance, jsonValue, value.converter, value.itemConverter,
-                        value.paramType, value.objectConverter,
-                        isDedup() ? new JsonPointerTracker(jsonPointer, jsonEntry.getKey()) : null, inType);
-                if (convertedValue != null) {
-                    setterMethod.write(t, convertedValue);
-                }
+            } catch (SetterMappingException alreadyHandled) {
+                throw alreadyHandled;
+            } catch (Exception e) {
```
rmannibucau, line 440: Shouldnt it let most RuntimeException go to not break exception handling in user code?

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/ExceptionMessages.java
```diff
@@ -0,0 +1,77 @@
...
+import static javax.json.JsonValue.ValueType.OBJECT;
+import static javax.json.JsonValue.ValueType.STRING;
+import static javax.json.JsonValue.ValueType.TRUE;
+
+public class ExceptionMessages {
+
+    private ExceptionMessages() {
+    }
+
+    public static String simpleName(final Type type) {
```
rmannibucau, line 35: Cant we use getTypeName instead to simplify it?

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/ExceptionMessages.java
```diff
@@ -0,0 +1,77 @@
...
+import java.lang.reflect.Type;
+
+import static javax.json.JsonValue.ValueType.ARRAY;
+import static javax.json.JsonValue.ValueType.FALSE;
+import static javax.json.JsonValue.ValueType.NUMBER;
+import static javax.json.JsonValue.ValueType.OBJECT;
+import static javax.json.JsonValue.ValueType.STRING;
+import static javax.json.JsonValue.ValueType.TRUE;
+
+public class ExceptionMessages {
```
rmannibucau, line 30: Final ;)

rmannibucau (changes requested):
Few picky changes but it is perfect functionally

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
dblevins: I know internally we throw a lot of RuntimeExceptions like IllegalArgumentException and the NullPointerException I fixed still got wrapped.  To be extra sure I just updated #90 with a new test case that checks how we handle it when users throw a RuntimeException from their setters.  Looks like we always wrap their RuntimeException with JsonbException and generate a new message.

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/ExceptionMessages.java
dblevins: I mentioned it a bit in JOHNZON-370 expecting feedback here.  My personal preference as a user is the short names of the classes involved as it greatly helps me "see" the exception text.  When the full packages are involved I am a slow reader so I start "scanning forward" and inevitably miss important exception text and it just takes me way longer and I find it overall more painful than helpful.

That said, the more detailed exception text that follows always does list the full package and class names and I intentionally left them long and didn't "fix" them.  I felt that was a good balance.  The first line is quick and short and filled with enough clues to be useful for normal cases.  The second line can help you in rare cases where a tiny bit more info is helpful.

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/ExceptionMessages.java
dblevins: Done!

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
rmannibucau: I know but in mapper we let runtime go.
More than happy if we subclass java.lang.XException to catch the cases you mentionned but user exceptions (converter, adapter, ...) nor johnzon exception unrelated to that handling should be wrapped IMHO.

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/ExceptionMessages.java
rmannibucau: Simple name is ok for me and a substring, just having a simpler name extraction with more caseq handling is nicer from a code standpoint IMHO.

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
dblevins: Ok, to make sure I understand correctly, given code like the following (from the website):

```
final Mapper mapper = new MapperBuilder().build();
final MySuperObject otherObject = mapper.readObject(inputStream, MySuperObject.class);
```
If MySuperObject throws a RuntimeException from its setter or constructor that exact exception instance can be caught if `readObject` was wrapped in a try-catch.  Is that basically right?

dblevins: I created a test in PR #90 in the `johnzon-mapper` module that doesn't use JSON-B and instead uses `Mapper.readObject`.  The handling seems identical and any RuntimeException thrown from the user's code is wrapped in a `MappingException`.  If there is still a scenario I'm missing, the most helpful way forward would be if you hack up a small test case I can use as a reference.

rmannibucau: @dblevins so you mean you didn't change the exception stacktrace at all? If so fine for me (indeed it includes `getCause()` ;)). It is indeed about setters+getters (`write`/`read`) but also `Adapter`/`Converter`s (`toValue`). Think we should limit our breaking change to subclass existing exception in our own codebase but not force users to change their specific exception handling. A quick option is to refine the try/catch block to only needed parts I guess and passthrough `RuntimeException` as this which should enable to handle the cases you want to catch without side effects and other changes.

dblevins: A compassionate request to not use ";)" while we work this out as I find the explanations unclear and it makes me feel teased.

On the code I dig deeper as to how a user can find their own RuntimeException on a `mapper.readObject` call:

 - Thrown from a Converter: `catch (MapperException e) { e.getCause(); }` will return the RuntimeException
 - Thrown from a setter: `catch (MapperException e) { e..getCause().getCause(); }` will return the RuntimeException
 - Thrown from a constructor: `catch (IllegalStateException e) { e.getCause(); }` will return the RuntimeException

With all the scenarios handled very differently, I'm not sure I see a clear path forward that doesn't involve some kind of change.

It seems at the very least the third scenario should be fixed so that users can catch MapperException, bringing it in line with the other two.  Beyond that, I'm not sure we want to make strict guarantees about what exception a user does or does not find when they call `getCause()` In concrete terms that would greatly reduce the kind of changes we can make to the internals.

The most pragmatic perspective I can imagine is that we simply guarantee their exception will be in the `getCause()` chain and they can interate through it to find it, but should not expect it to be at a specific position.

Thoughts?

For visibility, I have till Friday, then it's a couple weeks before I can do more.  We may need to break this into two goals; tweaks if any to get this PR in; possible longer-term changes.

One very simple option could be we simply make the next release 1.3 instead of 1.2.19 and say, "it's different, deal with it"

rmannibucau: Was thinking the same about the versioning but the impact is that we should probably do it for all exceptions to not look half baked - this is why i proposed to stick to a backward compat option and refine the try/catch in nested calls and not around the whole call  but Im happy with the 1.3 one if coverage is reached too.

To refine my qtatement, issue is not having directly or not the exception but more about not rewrapping it in a minor version. Concretely if today i know my cause is the 3rd cause then it must not become the 4th in 1.2.19 - ok in 1.3. similarly for the type - cause due to jsonb abstraction, to stay portable, classname is sometimes checked.

Still thinking out loud but think we should just lean toward 1.3 here to ease things for everyones, wdyt?

dblevins: I think it's very fair to be concerned about people's exception handling in a patch release, no matter how brittle.  I'd probably lean toward the 1.3 as well so we had the flexibility to make changes we feel are the best changes.  What kind of coverage do we think we'd want for a 1.3?  Happy to hack up a serious number of tests :)  I just need a list of scenarios to work from.

rmannibucau: I'd say getters too even if here we should probably invent something or just say we accept tostring usage (think it is not good but can we do much better?). Can need an api review (basically each time user code is used it can happen so starting from the API enables to be exhaustive without trying to do it from the code - it is what I'd do at least).

dblevins: I'm not sure I understood the last sentence.  When you say through API review without trying to do it in code, what do you have in mind?

rmannibucau: Checking jsonb+mapper user api you get all user extensions points outside getters/setters/constructors (thinking to adapters, serializers etc).

dblevins: Cool.  I'll bring the Mapper API tests from PR #90 into this PR expand everything to cover the getters.  I have tests for the constructor for the bean, but now that we're talking I likely need to add tests for the constructor of the converter.  Is there a particular test or doc that would help me see how adapters and serializers work?

In these tests do we want to verify the exact position of the exception in the cause-chain or do we want to simply verify it is in the chain (i.e. not lost/swallowed)?

rmannibucau: If we target 1.3 I would let you choose, if we target 1.2.x we should ensure position are iso, and instances are user ones.

side note: maybe `@JsonbCreator` can be added for json-b too (factory method + constructor).

For mapper doc is there https://johnzon.apache.org/ (`@JohnzonConverter` but there are a few supported converter types - 3 IIRC), for jsonb it is the spec. If you need there should be test for all of these proprietary api.

dblevins: Let's target 1.3 and take the stance that users can guarantee their exception will be in the chain, but not necessarily where.  I'll write tests that ensure the exact user's exception instance is present in 100% of the covered cases and work to get coverage for literally everything I can find.

Adding @JsonbCreator to the list.  Is there a Mapper equivalent to watch out for?

rmannibucau: @dblevins not for factry methods, for constructor there was `ConstructorProperties` (to select one of the constructors if no no-arg one exist) and there is JsonbAccessMode programmatic hook (Factory) which enables any impl of a factory but no annotation.

dblevins:
@rmannibucau we should consider merging this PR and working on any further exception handling fixes in separate PRs.  My concern is this PR is getting a bit too large for people to digest.  If we start fixing more API exception handling in this PR it's going to get way bigger.  We may start seeing knock-on impact to other tests and may want to rethink, rework or abandon some of those fixes.

At this moment there is just one more improvement in exception handling over what was previously added (JOHNZON-382), the rest is test coverage for the way things are, with no further change in behavior.

Thoughts?

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
```diff
@@ -373,11 +377,20 @@ private Object buildObject(final Type inType, final JsonObject object, final boo
...
-        } else {
-            t = classMapping.factory.create(createParameters(classMapping, object, jsonPointer));
+        try {
+            if (classMapping.factory.getParameterTypes() == null || classMapping.factory.getParameterTypes().length == 0) {
+                t = classMapping.factory.create(null);
+            } else {
+                t = classMapping.factory.create(createParameters(classMapping, object, jsonPointer));
+            }
+        } catch (Exception e) {
+            final String message = String.format("%s cannot be constructed to deserialize %s: %s%n%s",
```
rmannibucau, line 387: can it be a mapper exception (normally yes from create parameter
do we want to bypass this catch adding a catch(MapperException) before?

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
```diff
@@ -399,36 +412,43 @@ private Object buildObject(final Type inType, final JsonObject object, final boo
...
+                    }
                 }
-                final Object convertedValue = toValue(
-                        existingInstance, jsonValue, value.converter, value.itemConverter,
-                        value.paramType, value.objectConverter,
-                        isDedup() ? new JsonPointerTracker(jsonPointer, jsonEntry.getKey()) : null, inType);
-                if (convertedValue != null) {
-                    setterMethod.write(t, convertedValue);
-                }
+            } catch (SetterMappingException alreadyHandled) {
```
rmannibucau, line 447: should it be MapperException directly?

rmannibucau:
@dblevins any hope you review the question about catch(MapperException) (2 locations)? Except that where we potentially can avoid a double MapperException (e+cause) it is ready to merge for me.

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
dblevins: I'm my testing I haven't seen any scenarios where MappingException is thrown.   Issues appear to get wrapped as IllegalStateException before reaching this catch.

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
dblevins: If we change it to catch and not wrap MapperExcpetion, than we defeat the goal of the PR as now we'll no longer get the additional context of what class, field name, field type and json snippet involved.  The MapperExcpetions that happen downstream of this catch are accurate, but often don't provide enough context to be very useful in practice.  So we catch those exceptions and wrap them to provide the class, field, field type, and json snippet.  We don't, however, want to do that recursively so we let exceptions of just this type go by.

I did consider other means to collect that information and make it available.  My first instinct was as having an object to aggregate that kind of state that we could pass everywhere.  It occurred to me that all that information was already in the java stack, being tracked exactly the way we needed, we just needed to find the right place to catch and add that detail.  What I liked about that is this kind of tracking is really only useful for exceptions, but if we took a tracking-object kind of approach we'd be paying for it all the time, exception or not.  I decided it was best not to let myself over-engineer it and take advantage of what is already in the stack.

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
rmannibucau: Hmm, this can need to be refined if we get 2 different mapper exceptions in the stack and avoid this ISE (in particular cause we go 1.3 now).

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
rmannibucau: Did you consider making MapperException mutable - overriding getMessage and enabling to inject/set the context? Sounds like a good compromise to me avoiding huge stacks with no context gain.

dblevins: Is it possible we can move forward with this and evaluate that for future releases?

rmannibucau: Well we decided to move to 1.3 to refine exception model so whereas Im fine to merge as this, I guess it is sufficient to hold a release too.

dblevins: That's great.  I really appreciate it.

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingParserImpl.java
dblevins: Pushed something that should prevent us from doing this wrapping more than once

rmannibucau:
@dblevins FYI here is a test showing the comment I made in the PR to explain why current exception handling is not yet release friendly/ready (and why I spoke of mutable exception as an option - note that I don't care instantiating 3-4 instances but I care about instantiating 2 stack traces since they are often very costly in apps so a MapperExceptionBuilder which would be a RuntimeException we catch internally and `.build()` to a MapperException after having saved the context in the `Mapper` instance or alike would perfectly work if you don't want to make `MapperException` mutable on user side - which is a very fair point): https://github.com/apache/johnzon/commit/c2bf6945c2ca43befa93e15d821e0dee439aa8de#diff-9a970ccae72ea07cf4f226b1066a425fda9d97f48e06f1bc6f12079da28ca2f4R38

PR #93, opened 2022-10-01
make build reproducible

hboutemy:


SingingBush:
@hboutemy could you rebase this on top of the recent changes in master along with also putting `<maven.compiler.target>1.8</maven.compiler.target>` into the properties of the main pom because the apache super pom set's it to 1.7

@rmannibucau please can this get merged, it'll help with my next PR

hboutemy:
@SingingBush rebase done
@rmannibucau ok to merge?

rmannibucau:
Ok for me, i cna handle the fact to keep a dynamic timestamp for snapshot after the merge so let's go

hboutemy:
>  I can handle the fact to keep a dynamic timestamp for snapshot after the merge

what does that mean?

rmannibucau:
Maybe to only set a build timestamp in release profile -  maybe based on the version, didnt think of that last part.

hboutemy:
oh, you want a more complex setup: is there a concrete problem you're trying to solve?

if you really want to disable reproducible builds for SNAPSHOTs (for a reason I'd love to understand), you can create a profile and define a 1 char value: this will disable reproducible builds

rmannibucau:
It is more about having the real build timestamp in the jar than a meaningless value

rmannibucau:
Integrated, seems output does not have the pitfall I was thinking about cause project never set it up, thanks for the ready to use PR :+1:

PR #95, opened 2022-11-14
fixes for Javadoc linting errors

SingingBush:
small PR here for fixing some Javadoc linting problems

PR #96, opened 2022-11-14
update maven plugins

SingingBush:
building on the work done by @hboutemy in PR #93, and the javadoc fixes in PR #95, these changes fix some outstanding errors with Javadoc and aim to improve the build by making sure that maven plugins are defined in the root johnzon pom within the _pluginManagement_ section. 

In the case of the cobertura plugin, I've removed it entirely as it's been defunct for some time and Jacoco should be used instead. I've swapped the now dead findbugs plugin for spotbugs (more changes need to use v4 though). The checkstyle plugin needs further changes which can be in a future PR. I've also enabled javadoc linting as it no longer fails.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -22,10 +22,7 @@ jobs:
           distribution: ${{ matrix.dist }}
           cache: 'maven'
 
-      - name: Maven Package
+      - name: Maven Verify
         env:
           MAVEN_OPTS: -Dmaven.artifact.threads=64
-        run: mvn -V package --no-transfer-progress
-
-      #- name: Maven javadoc
-      #  run: mvn javadoc:javadoc -Ddoclint=all --no-transfer-progress
+        run: mvn -V verify --no-transfer-progress
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -22,7 +22,7 @@ jobs:
           distribution: ${{ matrix.dist }}
           cache: 'maven'
 
-      - name: Maven Verify
+      - name: Maven Package
         env:
           MAVEN_OPTS: -Dmaven.artifact.threads=64
-        run: mvn -V verify --no-transfer-progress
+        run: mvn -V package --no-transfer-progress
--- a/johnzon-jsonb/src/test/java/org/apache/johnzon/jsonb/JsonbRecordTest.java
+++ b/johnzon-jsonb/src/test/java/org/apache/johnzon/jsonb/JsonbRecordTest.java
@@ -34,18 +34,18 @@ public class JsonbRecordTest {
 
     @Test
     public void roundTrip() {
-        final JsonbRecord ref = new JsonbRecord(119, "Santa");
+        final Record ref = new Record(119, "Santa");
         final String expectedJson = "{\"_name\":\"Santa\",\"age\":119}";
         assertEquals(expectedJson, jsonb.toJson(ref));
-        assertEquals(ref, jsonb.fromJson(expectedJson, JsonbRecord.class));
+        assertEquals(ref, jsonb.fromJson(expectedJson, Record.class));
     }
 
     @JohnzonRecord
-    public static class JsonbRecord {
+    public static class Record {
         private final int age;
         private final String name;
 
-        public JsonbRecord(@JohnzonRecord.Name("age") final int age,
+        public Record(@JohnzonRecord.Name("age") final int age,
                            @JohnzonRecord.Name("name") @JsonbProperty("_name") final String name) { // simulate custom constructor
             this.age = age;
             this.name = name;
@@ -61,7 +61,7 @@ public class JsonbRecordTest {
 
         @Override
         public String toString() {
-            return "JsonbRecord{" +
+            return "Record{" +
                     "age=" + age +
                     ", name='" + name + '\'' +
                     '}';
@@ -75,8 +75,8 @@ public class JsonbRecordTest {
             if (o == null || getClass() != o.getClass()) {
                 return false;
             }
-            final JsonbRecord jsonbRecord = JsonbRecord.class.cast(o);
-            return age == jsonbRecord.age && Objects.equals(name, jsonbRecord.name);
+            final Record record = Record.class.cast(o);
+            return age == record.age && Objects.equals(name, record.name);
         }
 
         @Override
--- a/johnzon-jsonb/src/test/java/org/apache/johnzon/jsonb/JsonbRecordTest.java
+++ b/johnzon-jsonb/src/test/java/org/apache/johnzon/jsonb/JsonbRecordTest.java
@@ -34,19 +34,19 @@ public class JsonbRecordTest {
 
     @Test
     public void roundTrip() {
-        final Record ref = new Record(119, "Santa");
+        final JsonbRecord ref = new JsonbRecord(119, "Santa");
         final String expectedJson = "{\"_name\":\"Santa\",\"age\":119}";
         assertEquals(expectedJson, jsonb.toJson(ref));
-        assertEquals(ref, jsonb.fromJson(expectedJson, Record.class));
+        assertEquals(ref, jsonb.fromJson(expectedJson, JsonbRecord.class));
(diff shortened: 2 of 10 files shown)


.github/workflows/maven.yml
```diff
@@ -22,10 +22,7 @@ jobs:
...
 
-      - name: Maven Package
+      - name: Maven Verify
         env:
           MAVEN_OPTS: -Dmaven.artifact.threads=64
-        run: mvn -V package --no-transfer-progress
-
-      #- name: Maven javadoc
-      #  run: mvn javadoc:javadoc -Ddoclint=all --no-transfer-progress
+        run: mvn -V verify --no-transfer-progress
```
rmannibucau, line 28: we don't need verify AFAIK so let's not misguide users/us with an irrelevant command in the project

johnzon-jsonb/src/test/java/org/apache/johnzon/jsonb/JsonbRecordTest.java
```diff
@@ -34,19 +34,19 @@ public class JsonbRecordTest {
...
+        final JsonbRecord ref = new JsonbRecord(119, "Santa");
         final String expectedJson = "{\"_name\":\"Santa\",\"age\":119}";
         assertEquals(expectedJson, jsonb.toJson(ref));
-        assertEquals(ref, jsonb.fromJson(expectedJson, Record.class));
+        assertEquals(ref, jsonb.fromJson(expectedJson, JsonbRecord.class));
     }
 
     @JohnzonRecord
-    public static class Record {
+    public static class JsonbRecord {
```
rmannibucau, line 44: why was it needed?

johnzon-osgi/pom.xml
```diff
@@ -29,6 +29,10 @@
   <name>Johnzon :: Support for OSGI Jaxrs Whiteboard</name>
   <packaging>bundle</packaging>
 
+  <properties>
+    <maven.javadoc.skip>true</maven.javadoc.skip>
```
rmannibucau, line 33: guess it will break the release no?

pom.xml
```diff
@@ -103,7 +116,7 @@
         <plugin>
           <groupId>org.apache.maven.plugins</groupId>
           <artifactId>maven-shade-plugin</artifactId>
-          <version>3.2.3</version>
+          <!-- Version defined in Apache parent pom -->
```
rmannibucau, line 119: let's keep it there to not be unstable due to "external" parent pom

pom.xml
```diff
@@ -164,16 +177,50 @@
...
+        </plugin>
       </plugins>
     </pluginManagement>
+
     <plugins>
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
         <configuration>
-          <source>${java-compile.version}</source>
```
rmannibucau, line 174: why? should stay in the project IMHO

pom.xml
```diff
@@ -239,7 +286,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-site-plugin</artifactId>
-        <version>3.5.1</version>
```
rmannibucau, line 242: same

pom.xml
```diff
@@ -375,7 +421,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-javadoc-plugin</artifactId>
-        <version>3.2.0</version>
+        <!-- Version defined in Apache parent pom -->
```
rmannibucau, line 424: same

pom.xml
```diff
@@ -387,8 +433,7 @@
               <detectJavaApiLink>false</detectJavaApiLink>
               <detectLinks>false</detectLinks>
               <detectOfflineLinks>false</detectOfflineLinks>
-              <doclint>none</doclint>
-              <source>8</source>
```
rmannibucau, line 391: source needed I think

pom.xml
```diff
@@ -397,7 +442,7 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-surefire-plugin</artifactId>
-        <version>3.0.0-M3</version>
+        <!-- Version defined in Apache parent pom -->
```
rmannibucau, line 445: same but can be upgraded ;)

pom.xml
```diff
@@ -470,25 +494,7 @@
...
-                  <version>[3.3,)</version>
-                </requireMavenVersion>
-                <requireJavaVersion>
-                  <version>[${java-compile.version},)</version>
-                </requireJavaVersion>
-              </rules>
-            </configuration>
-          </execution>
-        </executions>
+        <!-- Version (and configuration) defined in Apache parent pom -->
```
rmannibucau, line 497: same

pom.xml
```diff
@@ -517,12 +523,11 @@
       <plugin>
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-site-plugin</artifactId>
-        <version>3.5.1</version>
+        <!-- Version defined in Apache parent pom -->
```
rmannibucau, line 526: same

rmannibucau (changes requested):
think we shouldnt rely on apache parent changes for key plugins and we should stick to package/install but not verify IMHO

rest looks very valuable to me

johnzon-jsonb/src/test/java/org/apache/johnzon/jsonb/JsonbRecordTest.java
rmannibucau: yes, was the intent - and to be honest it is not lowercase just to run well on later java versions ;) - cause it is supposed to mimic a `record`

.github/workflows/maven.yml
SingingBush: I thought verify was needed to ensure plugins like checkstyle and the javadoc linting kick in. If they are running as part of package then it can be reverted

johnzon-osgi/pom.xml
SingingBush: yes the osgi properties were causing a real problem for javadoc. I tried excluding all _package-info.java_ but didn't get it working so decided just to skip javadoc on that module

pom.xml
SingingBush: no problem, I'll go back through and specify the versions

pom.xml
rmannibucau: have to admit I prefer the configuration to be colocalised instead of split accross section, it enables a better readability in editors and avoids to scroll to get it (hover gives the value to be concrete). The other advantage for newbies is it avoids to know all conventions and check the doc/plugin.xml to get the defaults. wdyt?

pom.xml
SingingBush: Javadoc plugin will use _maven.compiler.source_, see: https://maven.apache.org/plugins/maven-javadoc-plugin/javadoc-mojo.html

.github/workflows/maven.yml
SingingBush: fixed in new commit

pom.xml
SingingBush: fixed in new commit. plugin versions are defined under _pluginManagement_ in main pom

pom.xml
rmannibucau: right but same than https://github.com/apache/johnzon/pull/96#discussion_r1021642715 (in particular using a parent which is not only ours like apache parent which sometimes make other choices than the defaults)

pom.xml
SingingBush: I've made a habit of specifying it in the property format on all maven projects I work on. Done that way for years and although there is some knowledge newcomers may need, it's something worth knowing, especially as that property is now used by numerous plugins (even non-apache). I can put it back to being explicit but would at least use the standard property key.

rmannibucau: it is ok for me to use the property with the default name, no issue there but ensure the wiring is explicit please (funnily I used the exact same reasonning so hope both can converge ;))

PR #98, opened 2023-01-12
WIP Jakarta EE 10 TCK setup

jeanouii:


rmannibucau:
While it does not break the build there is no issue, fine to merge with a property `tck.skip` (or alike) if it breaks the build by default and if it helps

jeanouii:
I can rebase it

jeanouii:
Running the suite locally to see how it looks

jungm:
From what I've seen so far both failing tests in `SpecificTypesMappingTest` (BigDecimal/BigInteger) are incredibly easy to fix - just remove the converters for those two types so they get passed as BigDecimal/BigInteger instead of as a String to jsonp

this brings me back to the good old mapper vs jsonb question: Do we break mapper "json compatibility" (json generated in mapper 1.x wouldn't be compatible with mapper 2.x) in favour of being jsonb compliant? Or Maybe sth that should be solved in jsonb only?

rmannibucau:
@jungm jsonb only while possible, that said we can also challenge the test since serializing numbers either as string or plain numbers (we cant do both with json spec and java impl) is not portable and likely considered as a bug by most consumers

jeanouii:
I'd say we don't break for breaking.

If we really need to break, now is the good/best time to do it because we already have a breaking change with Jakarta and a couple of other things. Better to do it now at once and not later.

Romain, can you elaborate?
I'm happy to challenge the test if I can get some context and more precise thoughts

jeanouii:
I rebased the PR and moved the TCK under a profile so we can merge this PR before it gets too big and keep working on the compliancy

PR #100, opened 2023-04-12
Implement JSON-B 3 Polymorphism

jungm:
related discussion: https://lists.apache.org/thread/vok6wprktky7qkvwk79x68gz6mjcp7rb


johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JohnzonBuilder.java
```diff
@@ -31,6 +31,7 @@
 import org.apache.johnzon.mapper.Mapper;
```
rmannibucau, line 31: guess except the setMappingsFactory the rest of the diff is no more needed?

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbPolymorphismHandler.java
```diff
@@ -0,0 +1,155 @@
...
+    }
+
+    /**
+     * Validates that only one parent class (superclass + interfaces) has {@link JsonbTypeInfo} annotation
+     *
+     * @param classToValidate class to validate
+     * @throws JsonbException validation failed
+     */
+    private void validateOnlyOneParentWithTypeInfo(Class<?> classToValidate) {
+        if (getParentClassesWithTypeInfo(classToValidate).size() > 1) {
```
rmannibucau, line 114: this looks called too often (at least twice per type if I'm not mistaken, maybe reuse the same instance for both needs?)

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MappingGeneratorImpl.java
```diff
@@ -179,23 +179,26 @@ public void doWriteObject(Object object, JsonGenerator generator, boolean writeB
...
+                    final Object result = classMapping.adapter.from(object);
+                    doWriteObject(result, generator, writeBody, ignoredProperties, jsonPointer);
+                    return;
                 }
+
                 if (writeBody) {
                     generator.writeStartObject();
                 }
-                final boolean writeEnd;
-                if (config.getSerializationPredicate() != null && config.getSerializationPredicate().test(objectClass)) {
```
rmannibucau, line 193: :+1:

johnzon-mapper/src/test/java/org/superbiz/ExtendMappingTest.java
```diff
@@ -62,9 +62,9 @@ public MyMappings() {
                     -1, true, true, true, false, false, false,
                     new FieldAccessMode(false, false),
                     StandardCharsets.UTF_8, String::compareTo, false, null, false, false,
-                    emptyMap(), true, false, true,
-                    null, null, null, null, null,
-                    type -> new EnumConverter(type)));
+                    emptyMap(), true, false, true, null,
```
rmannibucau, line 65: see previous comment but overall this test is a contract so any diff is likely a bug

src/site/markdown/index.md
```diff
@@ -497,57 +497,6 @@ in `MessageDecoder`.
     }
 
 
-
-### JSON-B Extra
```
rmannibucau, line 501: maybe keep it and explain:

1. very high level what is in the spec
2. how to migrate from extras module to the standard impl

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbPolymorphismHandler.java
```diff
@@ -0,0 +1,155 @@
...
+
+            List<Class<?>> parentClassesWithTypeInfo = getParentClassesWithTypeInfo(current);
+            current = parentClassesWithTypeInfo.isEmpty() ? null : parentClassesWithTypeInfo.get(0);
+        }
+
+        return entries;
+    }
+
+    public Class<?> getTypeToDeserialize(JsonObject jsonObject, Class<?> clazz) {
+        validateJsonbTypeInfo(clazz);
```
rmannibucau, line 83: this should be done one per class somehow since it is a runtime class (compared to getPolymorphismPropertiesToSerialize which is a classmapping init computation only) so maybe use ClassMapping to do the validation only one, can likely become a `Supplier<Class<?>>` using the supplier "pre" part to do the validation and `supplier#get` to impl the runtime only (current line 91+).

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbPolymorphismHandler.java
```diff
@@ -0,0 +1,155 @@
...
+        validateOnlyOneParentWithTypeInfo(clazz);
+
+        JsonbTypeInfo typeInfo = Meta.getDirectAnnotation(clazz, JsonbTypeInfo.class);
+        if (typeInfo == null || !jsonObject.containsKey(typeInfo.key())) {
+            return clazz;
+        }
+
+        JsonValue typeValue = jsonObject.get(typeInfo.key());
+        if (!(typeValue instanceof JsonString)) {
```
rmannibucau, line 92: `typeValue != null && typeValue.getValueType() != STRING` is faster due to JVM `instanceof` impl (to avoid when possible)

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbPolymorphismHandler.java
```diff
@@ -0,0 +1,155 @@
...
+            throw new JsonbException("Property '" + typeInfo.key() + "' isn't a String, resolving "
+                    + "JsonbSubtype is impossible");
+        }
+
+        String typeValueString = ((JsonString) typeValue).getString();
+        for (JsonbSubtype subtype : typeInfo.value()) {
```
rmannibucau, line 98: should be a map for runtime probably

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbPolymorphismHandler.java
```diff
@@ -0,0 +1,155 @@
...
+            }
+        }
+    }
+
+    /**
+     * Collects all parent classes (superclass + interfaces) that have the JsonbTypeInfo annotation
+     * @param clazz base class
+     * @return List of classes with JsonbTypeInfo annotation
+     */
+    private List<Class<?>> getParentClassesWithTypeInfo(Class<?> clazz) {
```
rmannibucau, line 147: wonder if the following pattern isnt more relevant (pseudo code):

```
final var superClass = clazz.getSuperclass();
if (superClass != null && hasTypeInfo(superClass)) {
  return superClass;
}
return Stream.of(clazz.getInterfaces()).filter(it -> hasTypeInfo(it)).findFirst().orElse(null);
```

Overall idea is to avoid to create a list to keep only the first element. For interface case it will not change a lot but for parent case it will make it way more efficient. wdyt?

rmannibucau (changes requested):
Did a full review now we agreed upon the design, it is mainly a few enhancements of the runtime and get back the user contract on some points but looks very close to be mergeable to me. Very good work and kudo to have enhanced the default polymorphism support which was not as great as we aim at!

tandraschko:
great work Markus, looking forward for this contribution!

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MapperConfig.java
```diff
@@ -98,6 +98,8 @@ public Object fromJson(JsonValue jsonObject, Type targetType, MappingParser pars
 
     private final SnippetFactory snippet;
 
+    private final Function<MapperConfig, Mappings> mappingsFactory;
+
     //CHECKSTYLE:OFF
     @Deprecated
     public MapperConfig(final LazyConverterMap adapters,
```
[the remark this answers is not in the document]
jungm: Is it okay to touch the other constructor of MapperConfig? (the one not used in ExtendMappingTest)
Or do I add a third one and mark the current non-deprecated one as deprecated?

src/site/markdown/index.md
jungm: don't know if migrating from jsonb-extras to jsonb3 will even be something users want to do - generated JSON format is different (`{"_type": "bla", "_value": {"someValue": 1}}` vs `{"_type": "bla", "someValue": 1}`)

Don't really want to add support for this format in jsonb module directly. It's out of spec and jsonb-extras was portable and is probably used on top of other jsonb impls
Maybe jsonb-extras shouldn't actually be deleted? It was pretty leightweight anyways

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/MapperConfig.java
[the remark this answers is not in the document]
rmannibucau: For the scope I have in mind it would be ok but it can be a limited scope (understand I will not fight if you modify the other but I have no real idea it can break other people).

src/site/markdown/index.md
rmannibucau: :), fine to revert the deletion, while the feature is there all good for me.

jungm: reverted the deletion + updated docs that extras polymorphism _shouldn't_ be used anymore

src/site/markdown/index.md
```diff
@@ -512,7 +512,9 @@ This module provides some extension to JSON-B.
 
 #### Polymorphism
 
-This extension provides a way to handle polymorphism:
+This extension shouldn't be used anymore if you don't absolutely rely on the JSON format it generates/parses.
```
rmannibucau, line 515: What about "If your aim is only to be polymorphic you can use the built-in JSON-B feature but some cases remain unhandled and this extension enable them (mainly JSON-B to JSON-B cases like configurations and service to service calls but unlikely javascript to java cases)".

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/Mappings.java
```diff
@@ -18,11 +18,16 @@
  */
 package org.apache.johnzon.mapper;
 
-import static java.util.Arrays.asList;
```
rmannibucau, line 21: oops (autoformat)?

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbPolymorphismHandler.java
```diff
@@ -0,0 +1,191 @@
...
+                }
+            }
+
+            current = getParentWithTypeInfo(current);
+        }
+
+        return result.toArray(Map.Entry[]::new);
+    }
+
+    public Class<?> getTypeToDeserialize(JsonObject jsonObject, Class<?> clazz) {
```
rmannibucau, line 76: implemented like that this must *not* be called at runtime, overall consider any reflection is forbidden at runtime except to discover the passed instance type so this impl should be limited to (pseudo code) `typeInfos.get(jsonObject.getString("..."))` so the reflection should have been stored in the handler earlier (likely in validate or getPolymorphismPropertiesToSerialize to avoid to do it > once) and the lookup shouldnt be a loop but a map access

rmannibucau (changes requested):
a small autoformat fix to do and `getTypeToDeserialize` to make runtime friendly and we are good to me

johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbPolymorphismHandler.java
jungm: reworked and reflection free now. Except for `Class#getName`, is using it okay? Or should we possibly completely move away from having the class as a parameter?

rmannibucau (approved):
Mappings diff can drop (autoformat?) then it looks good to me (early approval to avoid to need another pass just for style).
Very good work, thanks a lot for the effort!

jungm:
@rmannibucau dropped mappings diff

PR #103, opened 2023-04-25
don't serialize BigInteger/BigDecimal as strings in jsonb

jungm:
Fixes TCKs `SpecificTypesMappingTest.testBigDecimalMapping` and `SpecificTypesMappingTest.testBigIntegerMapping`

Breaks JSON compatibility between johnzon 1.x jsonb and johnzon 2.x jsonb when using BigInteger/BigDecimal without custom Adapters, imho users that need it should provider their own adapters

### DIFF
--- a/tck/jsonb/pom.xml
+++ b/tck/jsonb/pom.xml
@@ -137,6 +137,9 @@
           -->
           <systemProperties>
             <signature.sigTestClasspath>${project.build.outputDirectory}</signature.sigTestClasspath>
+
+            <johnzon.use-bigdecimal-stringadapter>false</johnzon.use-bigdecimal-stringadapter>
+            <johnzon.use-biginteger-stringadapter>false</johnzon.use-biginteger-stringadapter>
           </systemProperties>
           <argLine>-Duser.language=en -Duser.region=US</argLine>
         </configuration>


johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JohnzonBuilder.java
```diff
@@ -234,6 +236,10 @@ public Jsonb build() {
                         Integer.parseInt(it.toString()))
                 .ifPresent(builder::setSnippetMaxLength);
 
+        // drop johnzon-mapper BigInteger/BigDecimal built in adapters to not serialize those types as JsonString. See JSON-B spec 3.4.1
```
rmannibucau, line 239: 1. maybe replace the stored list by two toggles since we don't aim at storing it in mem at runtime there (this is used in a fallback of fallback we can rework later ;))
2. for me string is the right choice by default (see https://github.com/jakartaee/jsonb-api/issues/187) so maybe introduce a toggle in the jsonb config (and fallback on a system prop - static -  to easily pass tck?)

:+1: for te test

jungm: 1. reverted all changes I made in LazyConverterMap and simply added two booleans to enable/disable BigDecimal/Integer string adapters
2. Now defaults to string, though I do have mixed feelings about this one. Defaulting to string absolutely makes sense, but we're actively ignoring json-b spec here now

rmannibucau (approved):

jungm:
rebased onto `jakartaee-10-tck` and changed merge target to that branch so I can also set the system properties for jsonb TCKs

PR #105, opened 2023-05-10
fix(JOHNZON-397): protection against very slow BigDecimal to BigInteger conversion at large scale

jeanouii:
Requires maybe an option to configure the limit. Not sure if anyone needs something bigger than 1k digits.

### DIFF
--- a/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/util/JsonProviderUtil.java
+++ b/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/util/JsonProviderUtil.java
@@ -55,7 +55,9 @@ public final class JsonProviderUtil {
      * @param value the max big decimal scale to set on the provider
      */
     public static void setMaxBigDecimalScale(final JsonProvider provider, final int value) {
-        if (!"org.apache.johnzon.core.JsonProviderImpl".equals(provider.getClass().getName())) return;
+        if (!"org.apache.johnzon.core.JsonProviderImpl".equals(provider.getClass().getName())) {
+            return;
+        }
 
         try {
             SET_MAX_BIG_DECIMAL_SCALE.invoke(provider, value);


johnzon-core/src/main/java/org/apache/johnzon/core/JsonNumberImpl.java
```diff
@@ -117,4 +119,17 @@ private void checkFractionalPart() {
             throw new ArithmeticException("Not an int/long, use other value readers");
         }
     }
+
+    private void checkBigIntegerScale() {
```
rmannibucau, line 123: guess we need a toggle in the `Map` + `JsonbConfig` + `MapperBuilder` to disable it for backward compat - it stays legal?

johnzon-core/src/test/java/org/apache/johnzon/core/JsonNumberTest.java
```diff
@@ -62,20 +64,62 @@ public void equals() {
...
+
     @Test
     public void testBigInteger() {
-       
         JsonArray array = Json.createArrayBuilder().add(100.0200).build();
         Assert.assertEquals(new BigInteger("100"), array.getJsonNumber(0).bigIntegerValue());
+    }
+
+    @Test
+    public void testSlowBigIntegerConversion() {
```
rmannibucau, line 78: maybe test failing cases?

rmannibucau (changes requested):
Ok with a toggle + test for the error case to me

johnzon-core/src/test/java/org/apache/johnzon/core/JsonNumberTest.java
jeanouii: It's right bellow

jgallimore (approved):

PR #113, opened 2023-10-13
Site updates

jungm:
Fixes a few things I noticed while looking at the current site like broken links, javadoc not generating anymore, outdated download links, etc

Also Updates the site for JSON-P 2.1/JSON-B 3.0, documents properties I added in #103 and the removal of johnzon-jsonp-strict

### DIFF
--- a/src/site/markdown/index.md
+++ b/src/site/markdown/index.md
@@ -65,6 +65,8 @@ The generator factory supports the standard properties (pretty one for example) 
 
 ### JSON-P Strict Compliance (stable)
 
+**This has been removed with Johnzon 2.0.x, johnzon-core is now JSON-P compliant by default.**
+
 <pre class="prettyprint linenums"><![CDATA[
 <dependency>
   <groupId>org.apache.johnzon</groupId>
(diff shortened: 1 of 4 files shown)


src/site/markdown/index.md
```diff
@@ -353,6 +355,8 @@ JsonbConfig specific properties:
 * johnzon.accessMode: custom access mode, note that it can disable some JSON-B feature (annotations support).
 * johnzon.accessModeDelegate: delegate access mode used by JsonbAccessModel. Enables to enrich default access mode.
 * johnzon.failOnMissingCreatorValues: should the mapping fail when a `@JsonbCreator` misses some values.
+* johnzon.use-biginteger-stringadapter: Wether or not `BigInteger` is mapped as a string. `true` by default, set to `false` to ensure strict JSON-B 3 compliance
```
rzo1, line 358: Wether => Whether ?

rzo1 (approved):

rzo1 (approved):
lgtm. ping @jeanouii @rmannibucau

jungm:
But isn't EE 8 relocated to `jakarta.*` basically EE9?
the EE9 API jars will probably work fine with johnzon 2.0.x, but that also means users can't use new jsonp 2.1 and jsonb 3.0 features

rmannibucau:
@jungm well this is true but we still have this default issue (https://github.com/jakartaee/jsonp-api/blob/master/api/src/main/java/jakarta/json/spi/JsonProvider.java#L78 for jsonp for ex) so ultimately it can mean we still need custom jars (thinking a bit out loud but relying on the SPI is likely not what we want in maven and OSGi). Maybe a note on that issue of the spec can make it less a surprise to end user.

rmannibucau (approved):

PR #116, opened 2023-12-19
Add nashorn engine dependency in compiled scope for Java 15+

MathisGuillet1:
When trying to bump my application to Java 17, I faced issues with johnzon json-schema module.

As explained here https://stackoverflow.com/questions/71481562/use-javascript-scripting-engine-in-java-17 the nashorn javascript engine is removed starting jdk 15.

Here is the error I faced with JDK 17 :

`java.lang.NullPointerException: Cannot invoke "javax.script.ScriptEngine.createBindings()" because "org.apache.johnzon.jsonschema.regex.JavascriptRegex.ENGINE" is null`

Adding nashorn dependency in my application fixes the issue, but it is tricky to discover what gone wrong, so I suggest taking care of it here.

See here for the NullPointerException origin : https://github.com/apache/johnzon/blob/885cec0ef9484c3666c82315826143045f1c492a/johnzon-jsonschema/src/main/java/org/apache/johnzon/jsonschema/regex/JavascriptRegex.java#L32

### DIFF
--- a/johnzon-jsonschema/pom.xml
+++ b/johnzon-jsonschema/pom.xml
@@ -77,7 +77,6 @@
             <groupId>org.openjdk.nashorn</groupId>
             <artifactId>nashorn-core</artifactId>
             <version>15.3</version>
-            <scope>test</scope>
           </dependency>
         </dependencies>
     </profile>


jungm:
I think I'd prefer if johnzon would be more user friendly on this (explain in docs, and maybe log a warning/error when running on jdk >=15 and no ScriptEngine is available for js) instead of forcing a transient dependency like openjdk nashorn onto every user. They might be on an older jdk where this simply isn't needed or user prefers some other impl

another idea I have would be to make a "batteries included" johnzon-jsonschema-java15 module that has both johnzon-jsonschema and nashorn as dependencies. But I really doubt if this would be worth it 

wdyt? @rmannibucau @jeanouii

rmannibucau:
To be honest I always disable it to use java implementation, default is just more jsonschema friendly (java impl requires some constraints on the regex forms)

But clearly not an option to make it transitive by default, in particular cause there are other js impl (thinking to graal one) so we just respect the constract and user picks its impl if he uses js regexes.

hope it makes sense

MathisGuillet1:
Ok, let's close this, maybe just a log would help users with JDK15+ to figure out that they are missing an engine dependency

rmannibucau:
I'm fine moving `new ScriptEngineManager().getEngineByName("javascript");` in the constructor instead of static (does not bring much) and to throw an exception if null

PR #123, opened 2024-03-15
Guard Mapper.write against null writer

pingpingy1:
A `null` writer passed to the `Mapper.write` method, sensibly, throws an NPE upon flushing the buffer out. However, since the flushing occurs at a later time, debugging that this occurred is challenging. Thus, this commit implements a fail-fast `null` guard.

### DIFF
--- a/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/internal/DelegatingWriter.java
+++ b/johnzon-mapper/src/main/java/org/apache/johnzon/mapper/internal/DelegatingWriter.java
@@ -1,0 +1,79 @@
+/*
+ *     Licensed to the Apache Software Foundation (ASF) under one or more
+ *     contributor license agreements.  See the NOTICE file distributed with
+ *     this work for additional information regarding copyright ownership.
+ *     The ASF licenses this file to You under the Apache License, Version 2.0
+ *     (the "License"); you may not use this file except in compliance with
+ *     the License.  You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *     Unless required by applicable law or agreed to in writing, software
+ *     distributed under the License is distributed on an "AS IS" BASIS,
+ *     WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *     See the License for the specific language governing permissions and
+ *     limitations under the License.
+ */
+package org.apache.johnzon.mapper.internal;
+
+import java.io.IOException;
+import java.io.Writer;
+import java.util.Objects;
+
+public class DelegatingWriter extends Writer {
+    private Writer writer;
+
+    public DelegatingWriter(Writer writer) {
+        this.writer = Objects.requireNonNull(writer, "delegate writer must not be null");
+    }
+
+    @Override
+    public void write(final int c) throws IOException {
+        writer.write(c);
+    }
+
+    @Override
+    public void write(final char[] cbuf) throws IOException {
+        writer.write(cbuf);
+    }
+
+    @Override
+    public void write(final char[] cbuf, final int off, final int len) throws IOException {
+        writer.write(cbuf, off, len);
+    }
+
+    @Override
+    public void write(final String str) throws IOException {
+        writer.write(str);
+    }
+
+    @Override
+    public void write(final String str, final int off, final int len) throws IOException {
+        writer.write(str, off, len);
+    }
+
+    @Override
+    public Writer append(final CharSequence csq) throws IOException {
+        return writer.append(csq);
+    }
+
+    @Override
+    public Writer append(final CharSequence csq, final int start, final int end) throws IOException {
+        return writer.append(csq, start, end);
+    }
+
+    @Override
+    public Writer append(final char c) throws IOException {
+        return writer.append(c);
+    }
+
+    @Override
+    public void flush() throws IOException {
+        writer.flush();
+    }
+
+    @Override
+    public void close() throws IOException {
+        writer.close();
+    }
+}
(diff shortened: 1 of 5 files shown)


jungm:
We're already implicitly throwing NPEs when writer is null by invoking methods on it, don't think there is any need to explicitly guard for this

Your test is also already passing without modifying any other source code

pingpingy1:
@jungm Thanks for the response. I don't think I was quite thinking right when I wrote those tests.
The following is the NPE that I intended to address:

```java
public static void main(String args[]) throws Exception {
Mapper mapper = MapperLocatorDelegate.locate();
con_recv0.writeObject((Writer) null, (Object) "string");
}
```
```console
Exception in thread "main" java.lang.NullPointerException: Cannot invoke "java.io.Writer.write(char[], int, int)" because "this.val$from" is null
        at org.apache.johnzon.mapper.internal.Streams$1.write(Streams.java:45)
        at org.apache.johnzon.core.JsonGeneratorImpl.flushBuffer(JsonGeneratorImpl.java:481)
        at org.apache.johnzon.core.JsonGeneratorImpl.close(JsonGeneratorImpl.java:451)
        at org.apache.johnzon.mapper.Mapper.writeObject(Mapper.java:188)
        at Test.main(Test.java:17)
```

I know that the explicit `null` here makes the problem obvious, but if it weren't visible, the NPE is thrown not upon writing but later when flushing the buffer, which I thought could cause problems in some scenarios.

jungm:
Thanks, I understand a bit more what you want to accomplish now :slightly_smiling_face: 

Arguably `Cannot invoke "java.io.Writer.write(char[], int, int)" because "this.val$from" is null` can be a bit confusing (a user might think: What is this.val$from suppposed to be? What have I done wrong to make it null?")
I do believe renaming this parameter would be enough to clarify what has gone wrong though, adding null guards in the Mapper blows up the code with no real benefit. The exception is thrown immediately and easily debuggable, just not as clear why it is happening right now

You want to try going that path? Would also be an awesome opportunity to move the anonymous classes from `org.apache.johnzon.mapper.internal.Streams` into proper classes IMHO

rmannibucau:
+1 to just rename the parameter if it helps, having null there should have been identified way earlier in the chain anyway so shouldn't even be possible in terms of code flow. `requireNonNull` does not look required to me.

Side note: one required test is that it ends as a `JsonbException` in `johnzon-jsonb` and not a `NullPointerException` or anything else and that the runtime exception thrown in mapper stays of the same type for backward compat.

@jungm no big issue for me to make Streams anonymous classes inner classes but I'd like to avoid to make them surfacing as accessible or root classes, they are 100% dedicated to a particular internal need and should be droppable without caring about backward compat - was why it was done this way.

jungm:
Generally I'm not a friend of inner classes, yet alone anonymous classes - capturing variables from an "outside class" makes code super hard to read IMHO
I wouldn't worry too much about users assuming backwards compatibility with code that is literally in a package named `*.internal` tbh

rmannibucau:
@jungm maybe let do internal.DelegatingXStream classes (1-1 delegation pattern/decorator) and in `Stream` just override the close by a flush delegation, will enable to "expose" the delegating classes and maintain most of the code there and still prevent anyone to use the "skip the close" impl with an IDE magic trick.

> ....code that is literally in a package named *.internal...

it is not uncommon to see sun.internal in imports from time to time and then a "fix commit" and 99% of the time is it cause the IDE imported the wrong class and there it could be a big silent issue (leak with its impacts either in mem or ulimit).
So I prefer to not assume it.

> capturing variables from an "outside class" makes code super hard to read IMHO

guess we are not more at that stage, lambdas made it more than common ;).

pingpingy1:
@jungm @rmannibucau Not quite sure if I understand the delegation pattern, so I pushed a test commit that moves the anonymous `Writer` class to the `DelegatingWriter` class. Could you please check if this is what you intended?

johnzon-mapper/src/main/java/org/apache/johnzon/mapper/internal/DelegatingWriter.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public void flush() throws IOException {
+        writer.flush();
+    }
+
+    // Flushes instead of closing
+    @Override
+    public void close() throws IOException {
```
rmannibucau, line 78: (just explaining for this one but idea is the same for all types) here it should be `writer.*close()*`, then in `Streams` you do`return new DelegatingWriter(writer) { @Override public void close() {flush();} }`;
Idea was just to make most of the impl "standard" and just the specific part "hidden".

side note: even if I'm still not a fan of requireNonNull the minimum would be a valid error message (`writer = requireNonNull(writer, "delegate writer must not be null")`) but as mentionned in comment it is always wrong before in any code path so I would be interested to know how you got it and if the null check wouldn't sit better before in the codepath.

pingpingy1: I got what you mean. Let me make those changes and notify you then.

pingpingy1:
@rmannibucau Implemented all four delegated classes! PTAL

jungm (approved):
LGTM

only thing we can still discuss is if we want to make the DelegatingX constructors protected to ensure nobody accidentally uses these internal classes

rmannibucau:
Ok for me, to be picky (not a blocker) I'd like to get the param and field `final` in the delegating classes and `// no-op` in the payload instead of aside the method (an autoreformat will break it I think in most IDE) but can be handled after the merge easily - I can even do it for you if you feel you did enough (I fully agree, thanks for that ;)).

pingpingy1:
@rmannibucau No worries! Just pushed those changes.

PR #127, opened 2024-04-23
reduce enum array allocation

sullis:
## Motivation

reduce enum array allocation

## Modifications

use cached enum .values() 

## Additional context

https://www.gamlor.info/wordpress/2017/08/javas-enum-values-hidden-allocations/

### DIFF
--- a/johnzon-core/src/main/java/org/apache/johnzon/core/JsonStreamParserImpl.java
+++ b/johnzon-core/src/main/java/org/apache/johnzon/core/JsonStreamParserImpl.java
@@ -384,8 +384,8 @@ public class JsonStreamParserImpl extends JohnzonJsonParserImpl implements JsonC
 
     @Override
     public Event currentEvent() {
-        return previousEvent >= 0 && previousEvent < Event.values().length
-                ? Event.values()[previousEvent]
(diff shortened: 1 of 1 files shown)


johnzon-core/src/main/java/org/apache/johnzon/core/JsonStreamParserImpl.java
```diff
@@ -34,6 +34,8 @@
 
 //This class represents either the Json tokenizer and the Json parser.
 public class JsonStreamParserImpl extends JohnzonJsonParserImpl implements JsonChars {
+    private static final Event[] EVENT_VALUES = Event.values();
```
jungm, line 37: No need to define this here again, in `org.apache.johnzon.core.JsonCahrs` there's already a `Event[] EVT_MAP =Event.values()`

jungm (changes requested):
Nice catch! There's a tiny detail I noticed, when it's resolved I think we can merge this

johnzon-core/src/main/java/org/apache/johnzon/core/JsonStreamParserImpl.java
sullis: Fixed.

PR #128, opened 2024-06-24
[JOHNZON-404] don't rely on jdk internals for BoundedOutputStreamWriter

jungm:


rmannibucau (commented):
A few questions:

1. why don't you need to handle overflow (can be a reason but it is surprising upfront)
2. the while loop can be done at once saving the loop time I think and I'd need to check but think the encoding as well
3. doesnt encoder already have a throw exception management?

guess we'll need a ton of tests or fallback on the outputstreamwriter othrrwise :(

jungm:
The idea is basically: while there's chars left try to encode them into the buffer. If the encoding overflows (it is indeed handled) flush the buffer to the underlying OutputStream. I ignored underflows because imo underflowing (no input data left) is just the desired state we want to end up in

To me [javadocs](https://docs.oracle.com/en%2Fjava%2Fjavase%2F11%2Fdocs%2Fapi%2F%2F/java.base/java/nio/charset/CharsetEncoder.html#encode(java.nio.CharBuffer,java.nio.ByteBuffer,boolean)) don't sound like CharsetEncoder throws when encoding, hence I added the throwException() call

Agree we probably need more tests, this is really just a first naive approach to move away from sun StreamEncoder

rmannibucau:
@jungm I understand that part but we are far from the implementation on java < 22 so pretty sure we'll find not backward compatible cases. What can be worth it is to bench oututstreamwriter+bufferedwriter on java 22 vs our impl on java 21, if not diff we just extend it for backward compat?

rmannibucau:
what we don't want is to have an "exploding" bugger, the internal buffer of `OutputStreamWriter` is acceptable (even if I agree undesired) but the facading buffer kind of saves things.

alternatively we can do an impl close to the old one of the jdk to just stay ~the same, wdyt?

jungm:
Did some more benchmarking (all ran on JDK 21 this time):

```
Benchmark                   Mode  Cnt   Score    Error   Units
MyBenchmark.buffered       thrpt   25  11.728 ±  0.042  ops/us
MyBenchmark.bufferedHuge   thrpt   25   0.277 ±  0.004  ops/us
MyBenchmark.bufferedSmall  thrpt   25  54.672 ±  0.328  ops/us
MyBenchmark.johnzon        thrpt   25   2.669 ±  0.022  ops/us
MyBenchmark.johnzonHuge    thrpt   25   0.029 ±  0.001  ops/us
MyBenchmark.johnzonSmall   thrpt   25  55.897 ±  0.188  ops/us
MyBenchmark.raw            thrpt   25  20.570 ±  0.431  ops/us
MyBenchmark.rawHuge        thrpt   25   0.278 ±  0.005  ops/us
MyBenchmark.rawSmall       thrpt   25  73.361 ±  0.273  ops/us
```

```java
@BenchmarkMode(Mode.Throughput)
@Timeout(time = 10)
@Warmup(iterations = 3)
@State(Scope.Benchmark)
@OutputTimeUnit(TimeUnit.MICROSECONDS)
public class MyBenchmark {
    private Writer johnzon;
    private Writer buffered;
    private Writer raw;

    private char[] dataSmall;
    private char[] data;
    private char[] dataHuge;

    @Setup
    public void setUp() {
        johnzon = new BoundedOutputStreamWriter(OutputStream.nullOutputStream(), StandardCharsets.UTF_8, 100);
        buffered = new BufferedWriter(new OutputStreamWriter(OutputStream.nullOutputStream(), StandardCharsets.UTF_8), 100);
        raw = new OutputStreamWriter(OutputStream.nullOutputStream(), StandardCharsets.UTF_8);

        dataSmall = "Hello World".toCharArray();
        data = "Hello world".repeat(100).toCharArray();
        dataHuge = "Hello world".repeat(10000).toCharArray();
    }

    @Benchmark
    public void johnzonSmall() throws IOException {
        johnzon.write(dataSmall);
    }

    @Benchmark
    public void johnzon() throws IOException {
        johnzon.write(data);
    }

    @Benchmark
    public void johnzonHuge() throws IOException {
        johnzon.write(dataHuge);
    }

    @Benchmark
    public void bufferedSmall() throws IOException {
        buffered.write(dataSmall);
    }

    @Benchmark
    public void buffered() throws IOException {
        buffered.write(data);
    }

    @Benchmark
    public void bufferedHuge() throws IOException {
        buffered.write(dataHuge);
    }

    @Benchmark
    public void rawSmall() throws IOException {
        raw.write(dataSmall);
    }

    @Benchmark
    public void raw() throws IOException {
        raw.write(data);
    }

    @Benchmark
    public void rawHuge() throws IOException {
        raw.write(dataHuge);
    }
}
```

BufferedWriter + OutputStreamWriter is almost always faster than BoundedOutputStreamWriter

rmannibucau:
Interesting.

You are totally right on the use case (basically whatever we do the tests should still pass without modifying them).

High level the use case is when you open a writer - typically from a server, http/ssh/... - and want to ensure some flushing happens often to avoid to hang if messages are small.

So thanks to your bench I think that mixing BufferedWriter + OutputStreamWriter with an autoflush would be insanely sufficient.

jungm:
For good measure, another benchmark with the new auto flushing BufferedWriter:
```
Benchmark                  Mode  Cnt   Score   Error   Units
MyBenchmark.johnzon       thrpt   25  12.111 ± 0.052  ops/us
MyBenchmark.johnzonHuge   thrpt   25   0.275 ± 0.005  ops/us
MyBenchmark.johnzonSmall  thrpt   25  54.425 ± 0.066  ops/us
```

No performance lost and does what we want 😉

jungm:
Really the only reason I flush before writing is to not break/alter the behavior we had in BoundedOutputStream before, [BoundedOutputStreamWriterTest](https://github.com/apache/johnzon/blob/master/johnzon-core/src/test/java/org/apache/johnzon/core/io/BoundedOutputStreamWriterTest.java#L43-L52) expects the data that didn't fit in the buffer in the last write operation to stay in the buffer and not be flushed immediately

PR #130, opened 2025-05-06
Tolerate null @JsonbCreator values when adapters are used

dblevins:
When using Field injection, null (missing) json values are naturally tolerated.

When using Constructor inject, null (missing) json values are explicitly tolerated in `MappingParserImpl.toObject`.  However, if an adapter was specified,  `MappingParserImpl.convertTo` would throw a null pointer as it did not have the same code to deal with null values.


dblevins:
"Missing json values fail when using @JsonbCreator and @JsonbTypeAdapter"
https://issues.apache.org/jira/browse/JOHNZON-413

rmannibucau (approved):

jungm (approved):

PR #131, opened 2025-05-13
Annotation symmetry tests

dblevins:
I noticed some differences in behavior on reading vs writing with JsonbTypeAdapters.  They appear to have very different code paths and some caching issues that can alter behavior if reads are done after writes.

These test attempt to verify:
  - behavior is the same for both reading and writing
  - consistent order of preference (overriding)

First round of tests cover JsonbTypeAdapter, JsonbTypeDeserializer and JsonbTypeSerializer.  Tests that fail are ignored so this can be a passing PR.  Inconsistencies are documented.

Don't have time now, but would like to return to this at some point in the future to cover JsonbDateFormat, JsonbNumberFormat and the four flavors of JohnzonConverter.  I would also like to test meta-annotation versions of the above.


dblevins:
Discoveries:
 - Reads fail with String JsonbTypeAdapter on deserialized class
    https://issues.apache.org/jira/browse/JOHNZON-414

 - Reads fail with Array JsonbTypeAdapter on deserialized class
 -

PR #136, opened 2026-04-08
Add JSONB config for useBigIntegerStringAdapter and useBigDecimalStri…

jgallimore:
…ngAdapter

Following on from https://github.com/apache/johnzon/pull/134, could this change be considered?

This should allow this:

```
final JsonbConfig config = new JsonbConfig().withPropertyNamingStrategy(PropertyNamingStrategy.IDENTITY)
          .setProperty("johnzon.useBigIntegerStringAdapter", "false")
          .setProperty("johnzon.useBigDecimalStringAdapter", "false")
          ;
```

PR #137, opened 2026-04-08
Add JSONB config for useBigIntegerStringAdapter and useBigDecimalStri…

jgallimore:
…ngAdapter

Following on from https://github.com/apache/johnzon/pull/134, could this change be considered?

This should allow this:

```
final JsonbConfig config = new JsonbConfig().withPropertyNamingStrategy(PropertyNamingStrategy.IDENTITY)
          .setProperty("johnzon.useBigIntegerStringAdapter", "false")
          .setProperty("johnzon.useBigDecimalStringAdapter", "false")
          ;
```

(backport of #136)


jeanouii:
For consistency, do you mind using the same code block than in the main branch so future user upgrades will still work.

PR #138, opened 2026-04-09
fix(JOHNZON-426): BigInteger/BigDecimal string adapter flags swapped

jgallimore:
Replacing #136. This incorporates this change: https://github.com/apache/johnzon/commit/dc6b20557259755b2448fe600673357770a2869a, and should be more consistent with master.


jeanouii (approved):
LGTM
Thank you Jon

PR #141, opened 2026-06-29
JOHNZON-431: Fix detection of class level annotations

ArneLimburg:


johnzon-jsonb/src/main/java/org/apache/johnzon/jsonb/JsonbAccessMode.java
```diff
@@ -1051,7 +1051,11 @@ private static <T extends Annotation> T getAnnotation(final Parameter param, fin
         if (annotation != null) {
             return annotation;
         }
-        return Meta.findMeta(param.getAnnotations(), api);
+        final T metaAnnotation = Meta.findMeta(param.getAnnotations(), api);
+        if (metaAnnotation != null) {
+            return metaAnnotation;
+        }
+        return Meta.getAnnotation(param.getType(), api);
```
rmannibucau, line 1062: should we filter api? this one is more specific for params so guess the fallback should be at higher level

ArneLimburg: What do you mean?

Note that the test is green, as soon, as I define a setter for "InnerObject" in "OuterObject" in the test.

Currently classes, that are only used as constructor parameters (and there is i.e. no setter with that class as parameter) are completely ignored by the rest of the code.

Grabbing them at a more general location would be somewhere deep in org.apache.johnzon.mappings.Mapper (probably in createClassMapping). That refactoring would have much more impact i.e. in scanning performance.

Shall I go that path (i.e. not only scanning fields, getters and setters there, but also constructor parameters)?

rmannibucau: My understanding of your change is that getAnnotation(parameter) does match a class annotation, this should go on getAnnotation(Type). Several annotations can't be set on types and since johnzon enables programmatic decoration this enables broken/undefined usages so I would fix the annotation read per annotation type and not in the parameter lookup method.

note that it is fine to add a getAnnotationOnParameterOrClass method to ease that but still called per api allowing it

hope it makes sense

ArneLimburg: Note that the getAnnotation(parameter) is private and just used at that location where the bug is.

So do you mean, I should just rename getAnnotation(...) to getAnnotationOnParameterOrParameterType(...) ?

rmannibucau: @ArneLimburg not sure what you mean, you mean it is a bug we can't read a `@JsonbProperty` from a type? my point was more to add a method in the spirit of `org.apache.johnzon.mapper.access.Meta#getIndirectAnnotation` which does "read at that location (param) else this one (type)" with meta annotation support and use it only for the annotations allowing it like `JsonbTypeAdapter`, `JsonbDateFormat`, `JsonbNumberFormat` and `JsonbTypeDeserializer` but not `JsonbProperty` and `JohnzonConverter` (they all use the same `getAnnotation`).

ArneLimburg: Now I got your point and directly implemented it. Feel free to have a look

rmannibucau (approved):

PR #144, opened 2026-07-15
Enable to control how record java model is mapped to JSON

rmannibucau:


rmannibucau:
where are we with this proposal? should I rebase it or close it - I'm fine with both but don't think the issue David pushed on the list is fixed on main? happy to get some feedback

